PR #27, opened 2023-08-08
Add continuous integration scripts with dependencies installed by apt

traversaro:
Add initial CI scripts, for now just installing the dependencies via apt . Note that I also added a step with that runs the tests, that now does not anything as no tests are defined.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -1,0 +1,96 @@
+name: C++ CI Workflow
+
+on:
+  push:
+  pull_request:
+  schedule:
+  # * is a special character in YAML so you have to quote this string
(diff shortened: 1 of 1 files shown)


traversaro:
The compilation works fine, now the only step that is failing is the install, as we do not have any install step. I opened https://github.com/robotology/gz-yarp-plugins/issues/28 to track this, in the meanwhile I can just disable the install step.

traversaro:
@GiacomoBisio @lucapa17 if you do not have experience with GitHub Actions, probably you can check some docs at https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions . Even if you do not understand everything, the import thing is that you check that the command used for compilation are the expected ones (note that I used Ninja generator to speed up the compilation).

PR #43, opened 2023-08-19
Replace Gazebo references with Gz

GiacomoBisio:


GiacomoBisio:
>  Similarly, we could add a namepsace to the class implementing the class, to go from `GzYarpCamera` to `gzyarp::Camera`. 
> 

@traversaro What do you mean with this?

traversaro:
> > Similarly, we could add a namepsace to the class implementing the class, to go from `GzYarpCamera` to `gzyarp::Camera`.
> 
> @traversaro What do you mean with this?

In general, when usng namespaces in C++ there are some good ideas to follow:
* Do not use `using namespace <..>`, unless in some cases (see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using)
* Do not place your code in the global namespace, but rather place it in a dedicated namespace

Regarding the second point, I suggested to put the code of this repo in a `gzyarp::` namespace. For an example limited to a single fine (that is not migrating all the user of the class) in https://github.com/robotology/gz-yarp-plugins/commit/ec5c35ed7d2d8a0d122c136b6119678795e2392d .

traversaro:
Great, it seems to be working fine! codecov is complaining as the error lines are not covered by tests, but I guess that is ok. If you think it is ok, I will also rename the repo in `gz-sim-yarp-plugins`, thanks!

traversaro (approved):

libraries/singleton-devices/Handler.cc
```diff
@@ -88,7 +88,7 @@ bool Handler::setDevice(std::string deviceDatabaseKey, yarp::dev::PolyDriver* de
         //device does not exists. Add to map
         if (!m_devicesMap.insert(std::pair<std::string, yarp::dev::PolyDriver*>(deviceDatabaseKey, device2add)).second) 
         {
-            yError() << " Error in GazeboYarpPlugins::Handler while inserting a new device pointer!";
+            yError() << " Error in gz-yarp-Plugins::Handler while inserting a new device pointer!";
```
traversaro, line 91: Can you review the error messages in this file? `gz-yarp-Plugins::Handler` sounds strange, probably we can just put namespace and name of the class?

lucapa17:
> > Great, it seems to be working fine! codecov is complaining as the error lines are not covered by tests, but I guess that is ok. If you think it is ok, I will also rename the repo in `gz-sim-yarp-plugins`, thanks!
> 
> Do you agree with the rename @lucapa17 @GiacomoBisio ?

Yes, we do.

PR #44, opened 2023-08-28
Add install step

### LINKED ISSUE #28: Add install step
We should calls to the [install CMake function](https://cmake.org/cmake/help/latest/command/install.html), and document on which environment variables to modify to ensure that Gazebo can find the installed plugins.

lucapa17:
Fix #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ sudo apt remove yarp && sudo apt autoremove
 ~~~
 mkdir build
 cd build
-cmake ..
+cmake -DCMAKE_INSTALL_PREFIX=<desired_install_prefix> ..
 make
-sudo make install
+make install
 ~~~
 To notify Gazebo of the new plugins compiled, it is necessary to modify the GZ_SIM_SYSTEM_PLUGIN_PATH environment variable:
 ```


libraries/singleton-devices/CMakeLists.txt
```diff
@@ -8,4 +8,9 @@ target_link_libraries(Handler
   PRIVATE
     ${YARP_LIBRARIES}
     YARP::YARP_os
-    YARP::YARP_init)
\ No newline at end of file
+    YARP::YARP_init)
+
+install(TARGETS Handler
+        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
```
traversaro, line 14: Two suggestions:
* We are using the `CMAKE_INSTALL_BINDIR` variable, that is defined by the `GNUInstallDirs` module (see https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), but we are not including anywhere the module (i.e. via `include(GNUInstallDirs)`). Probably everything works as that module is already included either by YARP or Gazebo config cmake, but to be sure it is better to include it (see for example https://github.com/robotology/how-to-export-cpp-library/blob/afb21efb655e7b1cc11636c3d42aad9e02fb626f/CMakeLists.txt#L32)
* As we are using CMake 3.16 as minimum cmake version, the `DESTINATION` variables have a default value that is the one that we are setting (see https://cmake.org/cmake/help/v3.16/command/install.html#installing-targets). So probably instead of listing all the `*** DESTINATION` parameters, we can just add `install(TARGET Handler)`.

README.md
```diff
@@ -35,10 +41,18 @@ sudo apt install yarp
...
 ```
 
-## Compilation
+## <a name="install_gz-sim-yarp-plugins"></a> Install gz-sim-yarp-plugins
 ~~~
 mkdir build
 cd build
 cmake ..
 make
+sudo make install
```
traversaro, line 50: In general it is not a good idea to install in `/usr/local` as it can quicky become messy and lead to unmantainable setups. I suggest instead to document:
```suggestion
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=<desired_install_prefix> ..
make
make install
```

and then reference /path/to/the/install/folder/lib in the following docs.

PR #70, opened 2024-01-25
Add clock plugin with basic functionality

### LINKED ISSUE #60: Port basic clock plugin functionality from gazebo-yarp-plugins
Clock plugin docs is https://github.com/robotology/gazebo-yarp-plugins/blob/master/plugins/clock/include/gazebo/Clock.hh#L35 .

The clock functionality are two:
 * - It publishes the simulation time on a Yarp port `/clock`
 * - It opens a Yarp RPC Port (`/clock/rpc`) to allows control of the simulation time.

This issue focus on the first functionality, i.e. publication of simulation time on yarp port /clock.

xela-95:
This PR adds to the gz-sim-yarp-plugins a clock plugin, with the only basic functionality of writing the current simulation time on the yarp `/clock` port.

Closes #60 , #73


plugins/clock/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+project(clock)
```
traversaro, line 1:
```suggestion
```

plugins/clock/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+add_library(gz-sim-yarp-clock-system SHARED Clock.cc)
+target_compile_features(gz-sim-yarp-clock-system PRIVATE cxx_std_17)
+
+target_link_libraries(gz-sim-yarp-clock-system
+  PRIVATE gz-plugin${GZ_PLUGIN_VER}::gz-plugin${GZ_PLUGIN_VER}
+  PRIVATE gz-sim${GZ_SIM_VER}::gz-sim${GZ_SIM_VER}
+  PRIVATE
+    ${YARP_LIBRARIES}
```
traversaro, line 10:
```suggestion
```

If you need more `YARP` targets, let's list them explicitly (probably `YARP::YARP_dev` and/or `YARP::YARP_sig` are necessary).

plugins/clock/CMakeLists.txt
xela-95: @traversaro we can remove this because it is a low level plugin that is not referenced in other CMakeLists?

plugins/clock/CMakeLists.txt
xela-95: @traversaro could you point me towards some reference in which to better understand this? I've seen in Yarp documentation the usage of `${YARP_LIBRARIES}`: https://yarp.it/latest/group__add__a__plugin.html

plugins/clock/CMakeLists.txt
traversaro: Not really. The main point is that `project` only make sense inside the root CMakeLists.txt of a project, and typically it contain project-specific metadata like the version and the programming languages used (see https://cmake.org/cmake/help/latest/command/project.html). In some specific cases, it could make sense to have it in a subdirectory, but in general it is only useless code that can create confusion.

plugins/clock/CMakeLists.txt
traversaro: Good point, in general this would need to be updated. The point is that you never actually know what the `${YARP_LIBRARIES}` variable will contain, as it depends on how `find_package(YARP)` has been called (with which arguments). So in so-called "Modern CMake" it is better to always use targets, that clearly identify a single library. In your case, just try to put `YARP::YARP_dev` in place of `${YARP_LIBRARIES}`, and if there are compilation error we can look into that.

xela-95: Solve in 8f6967faa26c75debd5b0443e3d0180ca9f9d55e

Putting YARP::YARP_dev in place of ${YARP_LIBRARIES} worked.

xela-95:
@traversaro I tried to develop a first basic unit test for the clock plugin, here: 

https://github.com/robotology/gz-sim-yarp-plugins/blob/4a06b08fa297f69f3959402f0ecd1d0dd23963c7/tests/clock/ClockTest.cc

The unit test instantiates a gazebo server fixture and reads the `/clock` port after a certain number of simulation steps have been ran. Then it checks whether the elapsed simulation time received on the port is the one expected (number of timesteps times delta T).

Next I will try to add test to check other expected behaviors, like the pause and reset of simulation.

plugins/clock/Clock.cc
```diff
@@ -20,35 +21,70 @@ using yarp::os::Network;
...
+        , m_lastTimestampSent(0)
     {
-        m_portName = "/clock";
-        m_clockPort.open(m_portName);
+        std::cout << "===========> constructor" << std::endl;
     }
 
     ~Clock()
     {
+        std::cout << "===========> destructor" << std::endl;
```
traversaro, line 37: When debugging, always use `std::cerr`, `std::cout` may be buffered (even if in theory with `std::endl` you flush it).

xela-95: Ok thank you!

xela-95:
In commit 5acde4f08c7066d4afbf20918ab7b7df9a16cac9 I've implemented the `ISystemReset` interface. In this way, every time the user requests a simulation reset through Gazebo UI this method is called instead of destructing and re-constructing the plugin with the consequent closing of the yarp port. Now for a subscriber to the port the reset is transparent and it will just see the simulation time starting back from zero.

Unfortunately I've been not able to implement a unit test since the Server object does not expose a `Reset()` method that I can use to inject such event.

Some issues and PRs talking about reset: 
- https://github.com/gazebosim/gz-sim/issues/1107
- https://github.com/gazebosim/gz-sim/issues/203#issuecomment-642138544

plugins/clock/Clock.cc
```diff
@@ -0,0 +1,91 @@
...
+            }
+            yInfo() << "Clock plugin initialized";
+        }
+    }
+
+    void PostUpdate(const UpdateInfo& _info, const EntityComponentManager& _ecm) override
+    {
+        if (_info.paused)
+        {
+            // yDebug() << "Simulation pause, skipping clock update";
```
traversaro, line 62:
```suggestion
            // yDebug() << "Simulation pause, skipping clock update";
```
Similarly, if a message is commented, we can just remove it from the code to keep code simple.

traversaro (approved):
Just a small comment regarding unnecessary debug messages. Beside that, I think it is good to go, remember to squash commits!

plugins/clock/Clock.cc
```diff
@@ -0,0 +1,91 @@
...
+{
+
+class Clock : public System, public ISystemConfigure, public ISystemPostUpdate, public ISystemReset
+{
+public:
+    Clock()
+        : m_portName("/clock")
+        , m_initialized(false)
+    {
+        yDebug() << "Clock plugin Constructor invoked";
```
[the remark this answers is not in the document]
xela-95: Oh sure, I thought these prints will be enabled by manually setting a higher level of verbosity, but actually I have seen them while debugging the code without doing anything. I will remove them.

Issue #75, opened 2024-01-29
CameraTest segfaults on apt CI, currently worked around my running test multiple time until it pass


traversaro:
See for example https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/7698197313/job/20976938788 .

xela-95:
Closed in #76

traversaro:
Actually I wanted to have an issue to track the underling issue of the CameraTest segfault in a non-deterministic way, but indeed probably an issue with a more appropriate name may be more indicated.

traversaro:
Thanks, I clarified the test.

Nicogene:
Hi!

I was developing the `DepthCameraTest` and I have encountered the same segmentation fault w/ apt dependencies (see ), I tried to get some info from valgrind and/or gdb, this is what I got after compiling both YARP and gz-sim-yarp-plugins in debug:
```

Thread 49 "DepthCameraTest" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffbb57c6c0 (LWP 104443)]
0x00007fffa0fdab10 in ?? ()
(gdb) bt
#0  0x00007fffa0fdab10 in ??? ()
#1  0x00007ffff743d340 in __GI___nptl_deallocate_tsd () at ./nptl/nptl_deallocate_tsd.c:73
#2  __GI___nptl_deallocate_tsd () at ./nptl/nptl_deallocate_tsd.c:22
#3  0x00007ffff7440890 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.
```

Nicogene:
I confirm that w/ `export LIBGL_ALWAYS_SOFTWARE=1` the test passes without crashing on closure!

```
14: [----------] Global test environment tear-down
14: [==========] 1 test from 1 test suite ran. (5407 ms total)
14: [  PASSED  ] 1 test.
1/1 Test #14: DepthCameraTest ..................   Passed    5.47 sec
```

traversaro:
Ack, then the problem was that one, on the one tracked here. For reference, an upstream related issue is https://github.com/microsoft/WSL/issues/11277 . If you want to use hardware acceleration but not nvidia due to this bug, you can also select the intel GPU by setting `export MESA_D3D12_DEFAULT_ADAPTER_NAME=Intel`, see https://github.com/microsoft/wslg/wiki/GPU-selection-in-WSLg .

PR #76, opened 2024-01-29
apt CI: Increase until-pass number to mitigate Camera test flakiness

traversaro:
See https://github.com/robotology/gz-sim-yarp-plugins/issues/75 . In general, it is better not to have red CI, otherwise they may hide valid regressions. So if some failure are know, better to suppress them rather then just keep a red CI.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -97,7 +97,8 @@ jobs:
       if: contains(matrix.build_type, 'Release')
       run: |
         cd build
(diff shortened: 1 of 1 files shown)


xela-95:
Yes I agree, thanks @traversaro

xela-95 (approved):

traversaro:
Feel free to merge, thanks!

xela-95:
> Feel free to merge, thanks!

Ok! Closes #75

PR #92, opened 2024-02-21
Add IPositionDirect interface implementation to ControlBoardDriver

### LINKED ISSUE #82: Add Position direct control to the control board plugin
We want to add a low level, high frequency position control, corresponding to the [Position Control Direct](https://www.yarp.it/latest/classyarp_1_1dev_1_1IPositionDirectRaw.html#details) present in [gazebo-yarp-plugins control board](https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h#L322-L328).

xela-95:
Closes #82

### DIFF
--- a/plugins/controlboard/include/ControlBoard.hh
+++ b/plugins/controlboard/include/ControlBoard.hh
@@ -89,6 +89,7 @@ private:
     double convertGazeboGainToUserGain(JointProperties& joint, double value);
     double convertGazeboToUser(JointProperties& joint, double value);
     double convertUserToGazebo(JointProperties& joint, double value);
+    bool setJointPositionLimits(const gz::sim::EntityComponentManager& ecm);
 };
 
 } // namespace gzyarp
--- a/plugins/controlboard/include/ControlBoard.hh
+++ b/plugins/controlboard/include/ControlBoard.hh
@@ -60,7 +60,7 @@ private:
     yarp::os::Network m_yarpNetwork;
     yarp::os::Property m_pluginParameters;
 
-    enum UnitsTypeEnum
+    enum class UnitsTypeEnum
     {
         METRIC = 0,
         SI = 1
--- a/plugins/controlboard/include/ControlBoard.hh
+++ b/plugins/controlboard/include/ControlBoard.hh
@@ -70,7 +70,7 @@ private:
     void updateSimTime(const gz::sim::v7::UpdateInfo& _info);
     bool readJointsMeasurements(const gz::sim::EntityComponentManager& _ecm);
     void checkForJointsHwFault();
-    bool updateReferences(gz::sim::EntityComponentManager& _ecm);
+    bool updateReferences(const gz::sim::UpdateInfo& _info, gz::sim::EntityComponentManager& _ecm);
     double getJointTorqueFromTransmittedWrench(const gz::sim::Joint& gzJoint,
                                                const gz::msgs::Wrench& wrench,
                                                const gz::sim::EntityComponentManager& ecm) const;
@@ -85,8 +85,10 @@ private:
                           std::vector<yarp::dev::Pid>& yarpPIDs,
                           size_t numberOfJoints);
     void setJointPositionPIDs(UnitsTypeEnum cUnits, const std::vector<yarp::dev::Pid>& yarpPIDs);
-    double convertUserGainToGazeboGain(int joint, double value);
-    double convertGazeboGainToUserGain(int joint, double value);
+    double convertUserGainToGazeboGain(JointProperties& joint, double value);
+    double convertGazeboGainToUserGain(JointProperties& joint, double value);
+    double convertGazeboToUser(JointProperties& joint, double value);
+    double convertUserToGazebo(JointProperties& joint, double value);
 };
 
 } // namespace gzyarp
--- a/plugins/controlboard/include/ControlBoard.hh
+++ b/plugins/controlboard/include/ControlBoard.hh
@@ -14,6 +14,7 @@
 #include <gz/sim/Types.hh>
 #include <sdf/Element.hh>
 
+#include <yarp/dev/ControlBoardPid.h>
 #include <yarp/dev/PolyDriver.h>
 #include <yarp/os/Network.h>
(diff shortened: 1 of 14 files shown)


xela-95:
> The codecov stuff is quite noisy, in a useless way from what I see. If you like @xela-95 we can remove it.

Agree, at least for now. From here they say how to disable codecov from commenting on the PR https://docs.codecov.com/docs/pull-request-comments#disable-comment

In the example they set this preference on a `codecov.yml` file, that we don't use. Maybe it's just a setting I can put into the CI step in which we trigger codecov?

xela-95:
> The codecov stuff is quite noisy, in a useless way from what I see. If you like @xela-95 we can remove it.

See #93

plugins/controlboard/include/ControlBoard.hh
```diff
@@ -59,14 +60,36 @@ private:
     yarp::os::Network m_yarpNetwork;
     yarp::os::Property m_pluginParameters;
 
+    enum class UnitsTypeEnum
+    {
+        METRIC = 0,
```
traversaro, line 65: What do you mean by "METRIC" ? How it is different from "SI" ? I thought they are equivalent?

xela-95: This is a distinction ported from the original plugin, see the [setPIDsForGroup_POSITION](https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriver.cpp#L1056-L1201) method. I think this property only means degrees vs. radians, since the only place where it is used is in that method where, based on that configuration parameter, the PID gains are converted from degrees to radians or not.

traversaro: Ahh, I see. The related code is https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriver.cpp#L1174-L1185 . Do you think this "metric" used to identify "degrees for angles" make sense? Can you check? If it does not make sense, probably we can use another word for meaning "degrees for angles"? We can still temporary support `metric` in the config files to simplify the Gazebo Classic --> gz-sim migration, but at least for the internal enum I would try to stick to something that make sense at least for us.

traversaro: To me, "metric" is just the ancestor of the International System of Units (see https://en.wikipedia.org/wiki/International_System_of_Units), so they are basically synonyms.

xela-95: Yes I agree entirely with your considerations; sorry for not pointing out this ambiguity while porting it. I thought the same as you, that metric and SI basically mean the same thing. Maybe for now we can change in the code the naming of variables to distinguish radian from degree angles, while maintaining the same configuration files keywords.

xela-95: I confirm that the `metric_units` configuration property is used in the three methods to set the PIDs for position, velocity and impedance controls:
- position: https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriver.cpp#L1056-L1201
- velocity:https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriver.cpp#L1202-L1313
- impedance: https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriver.cpp#L1314-L1397

xela-95: Done in https://github.com/robotology/gz-sim-yarp-plugins/pull/92/commits/751fdd296a5b5e5fd4194b8654f430b566407240

traversaro (approved):
Thanks! The codecov comments are still spammy and kind of make the rendering complex, anyhow this is not blocking. Remember to merge with squash or curate the commits, thanks!

xela-95:
> Thanks! The codecov comments are still spammy and kind of make the rendering complex, anyhow this is not blocking. Remember to merge with squash or curate the commits, thanks!

From the documentation https://docs.codecov.com/docs/github-checks#hiding-annotations-in-the-files-view, it says that you can hide all the annotations in that tab by pressing the "a" key. Otherwise, it should be possible to completely disable them from the CI yaml.

xela-95:
Merging!

traversaro:
> > Thanks! The codecov comments are still spammy and kind of make the rendering complex, anyhow this is not blocking. Remember to merge with squash or curate the commits, thanks!
> 
> From the documentation https://docs.codecov.com/docs/github-checks#hiding-annotations-in-the-files-view, it says that you can hide all the annotations in that tab by pressing the "a" key. Otherwise, it should be possible to completely disable them from the CI yaml.

Thanks, I did not know that. However, this needs to be done manually and for each file, so it is not really convenient. Is anyone checking those codecov annotations? If no one is checking those, what is the advantage of having them? In that case, I think we can disable them.

xela-95:
> > > Thanks! The codecov comments are still spammy and kind of make the rendering complex, anyhow this is not blocking. Remember to merge with squash or curate the commits, thanks!
> > 
> > 
> > From the documentation https://docs.codecov.com/docs/github-checks#hiding-annotations-in-the-files-view, it says that you can hide all the annotations in that tab by pressing the "a" key. Otherwise, it should be possible to completely disable them from the CI yaml.
> 
> Thanks, I did not know that. However, this needs to be done manually and for each file, so it is not really convenient. Is anyone checking those codecov annotations? If no one is checking those, what is the advantage of having them? In that case, I think we can disable them.

Agree, I just use the codecov check action to have a brief overview of the code coverage, not the single line annotations. I will open a PR to disable them.

Issue #103, opened 2024-03-01
Import ergoCub in Gazebo


xela-95:
With this issue we want to import one of the ergoCub models available in https://github.com/icub-tech-iit/ergocub-software into [Gazebo](https://gazebosim.org/home).

xela-95:
CC @traversaro

xela-95:
Habemus the first version of ergocub in Gazebo Harmonic! 
![image](https://github.com/robotology/gz-sim-yarp-plugins/assets/57228872/cc61debb-cbd3-43f6-ac5c-644c2d995587)

@traversaro I continued from the work you started on Friday, this time I:
- Converted the https://github.com/icub-tech-iit/ergocub-software/blob/dbbdd3940718d71ce5763999bfb857504cd772c5/urdf/ergoCub/robots/ergoCubGazeboV1_1/model.urdf to `sdf` format with `gz sdf -p model.urdf > model.sdf ` and commented out from the resulting sdf all the sensors;
- The sdf model is then included into another model that simply adds the fixed joint between world and root link:

 ```xml
<?xml version='1.0'?>
<sdf version='1.7'>
  <model name="ergocub_fixed">
    <include>
      <uri>model://ergoCub/robots/ergoCubGazeboV1_1</uri>
      <pose>0.0 0 1.0 0 0 3.14</pose>
    </include>

    <joint name="fixed_base" type="fixed">
      <parent>world</parent>
      <child>ergoCub::root_link</child>
    </joint>
  </model>
</sdf>
 ```
- This model is then loaded into a world file:

```xml
<?xml version="1.0"?>

<sdf version="1.7">
    <world name="turorial_controlboard">
        <physics name="1ms" type="ignored">
            <max_step_size>0.001</max_step_size>
            <real_time_factor>1.0</real_time_factor>
        </physics>
        <plugin
            filename="gz-sim-physics-system"
            name="gz::sim::systems::Physics">
        </plugin>
        <plugin
            filename="gz-sim-user-commands-system"
            name="gz::sim::systems::UserCommands">
        </plugin>
        <plugin
            filename="gz-sim-sensors-system"
            name="gz::sim::systems::Sensors">
        </plugin>
        <plugin
            filename="gz-sim-scene-broadcaster-system"
            name="gz::sim::systems::SceneBroadcaster">
        </plugin>

        <light type="directional" name="sun">
            <cast_shadows>true</cast_shadows>
[...]

Issue #102, opened 2024-02-28
Avoid race conditions in ControlBoardDriver


xela-95:
Many methods that `ControlBoardDriver` class implements have different overloads:
- Act on a single joint
- Act on all joints
- Act on a subset of joints

While for the first case the methods have a lock_guard to guarantee a safe access to control board data, the other two typically have not that, because in principle they call the method version for the single joint for every robot joint. However In many of these methods there still some control board data members are accessed without a lock, potentially causing race conditions.

An example is:

A way to avoid this is to switch to recursive mutex objects, so that the same mutex can be locked multiple times by nested functions.

Issue #90, opened 2024-02-20
Controlboard unit test failing on MacOs


xela-95:
The unit test developed in #86 revealed a potential error in how mutexes are managed in the plugins, as revealed in the unit test failing in the github CI for MacOs (see https://github.com/robotology/gz-sim-yarp-plugins/pull/86#issuecomment-1954286140).

In fact, this bug could be affecting all the plugins, not only the controlboard one, since the way in which they implement the singleton is the same.

xela-95:
Actually this is happening also for ConfigurationParsingTest. See https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/8263726863/job/22605783707

PR #109, opened 2024-03-05
Fix controlboard bug on resetPositionsAndTrajectoryGenerators at initial configuration

### LINKED ISSUE #108: Fix controlboard bug on `resetPositionsAndTrajectoryGenerators` at initial configuration
See https://github.com/robotology/gz-sim-yarp-plugins/issues/103#issuecomment-1978894636

xela-95:
Closes #108

### DIFF
--- a/.github/workflows/conda-forge.yml
+++ b/.github/workflows/conda-forge.yml
@@ -89,7 +89,7 @@ jobs:
       run: |
         cd build
         # ImuTest is skipped due to https://github.com/robotology/gz-sim-yarp-plugins/issues/54
-        ctest -E "^LaserTest|^CameraTest|^ImuTest|^ForceTorqueTest" --repeat until-pass:5 --output-on-failure -C ${{ matrix.build_type }} .
+        ctest -E "^LaserTest|^CameraTest|^ImuTest|^ForceTorqueTest|^ControlBoardCommonsTest" --repeat until-pass:5 --output-on-failure -C ${{ matrix.build_type }} .
 
     - name: Install
       run: |
--- a/plugins/controlboard/src/ControlBoard.cpp
+++ b/plugins/controlboard/src/ControlBoard.cpp
@@ -871,12 +871,15 @@ void ControlBoard::resetPositionsAndTrajectoryGenerators(gz::sim::EntityComponen
         {
             auto& joint = m_controlBoardData.joints.at(i);
             auto gzJoint = Joint(Model(m_modelEntity).JointByName(ecm, joint.name));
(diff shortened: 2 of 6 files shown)


traversaro:
Do you have any idea why codecov continues to spam messages?

xela-95:
> Do you have any idea why codecov continues to spam messages?

unfortunately my updates to the CI yaml were not effective... I think I need to create a separate `codecov.yaml` file to customize its behavior: https://docs.codecov.com/docs/codecov-yaml

xela-95:
Anyhow it is possible to reproduce the exception I got in https://github.com/robotology/gz-sim-yarp-plugins/issues/103#issuecomment-1978894636 by launching the single pendulum tutorial and by commenting out the `<initialConfiguration>` block: 
https://github.com/robotology/gz-sim-yarp-plugins/blob/1e86adb9247b159900c4eae312222950fa0f70f1/tutorial/single_pendulum/model.sdf#L99

Also here I verified that I had an empty vector as the result of calling:
 https://github.com/robotology/gz-sim-yarp-plugins/blob/98f78ae5fff2fbcc8f5f240b3cb87475cded5451/plugins/controlboard/src/ControlBoard.cpp#L874

xela-95:
We're getting (again) a failure in the macos tests, in particular on the new test implemented:
https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/8160658434/job/22307711533?pr=109#step:12:50

```
 libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

An interesting reference on SO: https://stackoverflow.com/questions/66773247/libcabi-dylib-terminating-with-uncaught-exception-of-type-std-1system-er

xela-95:
> We're getting (again) a failure in the macos tests, in particular on the new test implemented: https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/8160658434/job/22307711533?pr=109#step:12:50
> 
> ```
>  libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
> ```
> 
> An interesting reference on SO: https://stackoverflow.com/questions/66773247/libcabi-dylib-terminating-with-uncaught-exception-of-type-std-1system-er

Also this could be related to the bug #102, since different platforms I'mplement locking mechanisms in different ways, this problem maybe shows up only on macOs...

xela-95:
Since this macOS issue is not related to the current development and is already documented in #90, I will exclude the new test from it for the time being.

traversaro (approved):

xela-95:
Test passed! Merging 🚀

PR #110, opened 2024-03-06
Update codecov configuration

xela-95:
This PR:
- Introduces a `codecov.yml` file in which all the codecov configurations con be customized;
-  Updates codecov action to V4


xela-95:
In order to validate the `codecov.yml` file it is possible to invoke the following API: 
```
curl -X POST --data-binary @codecov.yml https://codecov.io/validate
```

and check if the API response is a 200 code

traversaro (approved):

PR #115, opened 2024-03-11
Remove relative includes by using `target_include_directories` and `target_link_libraries`

### LINKED ISSUE #114: Remove relative includes by using `target_include_directories` and `target_link_libraries`
A better way to handle includes in c++ is given by defining [target_include_directories](https://cmake.org/cmake/help/latest/command/target_include_directories.html) and [target_link_libraries](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) in each target's CMakeLists and then by including the necessary headers with <> brackets.

As an example, see https://github.com/robotology/gz-sim-yarp-plugins/blob/5d451a22e2b1a2e325eb17011a0be75cc466d323/plugins/controlboard

xela-95:
This PR affects the following plugins:
- base state
- camera
- force torque
- imu
- laser
- robot interface

The modifications it introduces for these plugins are the use of <> in includes (no more includes directive with relative paths) by configuring `target_include_directories` and `target_link_libraries` in a correct way for each plugin. It also updates names for the singleton classes of each plugin to be more explicative and avoid name clashes (e.g. robotinterface Handler and Imu Handler classes).

Closes #114


plugins/imu/IMU.cc
```diff
@@ -25,19 +25,19 @@ using namespace systems;
 namespace gzyarp
 {
 
-class IMU : public System,
+class Imu : public System,
```
traversaro, line 28: Not blocking, but just fyi `IMU` is a acronym (Inertial Measurement Unit), so that's the reason of why it was all capital. Anyhow, I seem that upstream `gz-sim` uses `Imu` style, so it is ok to name it like that.

traversaro (approved):

plugins/imu/IMU.cc
xela-95: Yep that was one of the reasons for which I updated the class name, if you prefer I can revert it to the previous version :)

traversaro: No no that is ok, I just wanted to understand what was happening, but I realized it while writing my comment.

xela-95:
Merging! 🚀

PR #126, opened 2024-03-19
Fix configuration of joint position limits from yarp configuration

xela-95:
Controlboard plugin has a bug since it does not handle joint position limits correctly when multiple joints are configured in a controlboard.

https://github.com/robotology/gz-sim-yarp-plugins/blob/954f5ca27350f3d747dbce8856f45ec008010bbe/plugins/controlboard/src/ControlBoard.cpp#L649-L661


traversaro (approved):

xela-95:
Merging! 🚀

PR #128, opened 2024-03-20
Fix timestamp written by clock plugin

xela-95:
This PR affects the clock plugin.

It fixes the format of the timestamp written on the `/clock` port to be what yarp requires.

### DIFF
--- a/plugins/clock/Clock.cc
+++ b/plugins/clock/Clock.cc
@@ -1,6 +1,8 @@
+#include <chrono>
 #include <gz/plugin/Register.hh>
 #include <gz/sim/System.hh>
 #include <gz/sim/Util.hh>
+
 #include <yarp/os/Bottle.h>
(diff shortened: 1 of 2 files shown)


traversaro:
Thanks, for reference the related code in gazebo-yarp-plugins and YARP:
* https://github.com/robotology/gazebo-yarp-plugins/blob/bad19ba865278d38b7e93b31706e42ff487b0bb7/plugins/clock/src/Clock.cc#L167
* https://github.com/robotology/yarp/blob/2e513a64f2ffaa43b91c212c4bfbf20049b576b7/src/libYARP_os/src/yarp/os/NetworkClock.cpp#L103-L105

traversaro (approved):

xela-95:
xref to the issue with NetworkClock: https://github.com/robotology/yarp/issues/3096

xela-95:
Merging! 🚀

PR #139, opened 2024-04-03
Use gtest installed from package managers

traversaro:


traversaro:
@xela-95 if it is ok, feel free to merge with squash as commits are not really curated.

xela-95 (approved):
Thank you @traversaro!

PR #157, opened 2024-04-10
Update installation instructions

### LINKED ISSUE #156: Update installation instructions
Now that we have gz-sim-yarp-plugins in conda-forge and robotology-superbuild, probably we can change the installation section in the README (https://github.com/robotology/gz-sim-yarp-plugins?tab=readme-ov-file#installation) to refer to them, see for example https://github.com/robotology/idyntree?tab=readme-ov-file#installation (or anything else you find more user friendly). It is worth to keep the installation from source instructions, but I would keep them in a separate page to avoid confusing new users.

xela-95:
Closes #156

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,114 +14,41 @@
 
 ## Installation
 
-At the moment we do not provide any binary for `gz-sim-yarp-plugins`, so you need to compile it from source, either
-installing the dependencies with conda-forge on Linux, macOS or Windows, or apt on Ubuntu.
+### conda (recommended)
 
-### Compile from source using conda-forge dependencies on Linux, macOS or Windows
+You can easily install the C++ and Python library with via [`conda-forge`](https://conda-forge.org) using the following command:
 
-Create and activate an environment with the required dependencies:
+```bash
+conda install -c conda-forge gz-sim-yarp-plugins
+```
 
-~~~
-mamba create -c conda-forge -n gsypdev libgz-sim8 yarp ycm-cmake-modules cmake ninja pkg-config cmake compilers gtest
-mamba activate gsypdev
-~~~
+If you are not familiar with conda or conda-forge, you can read an introduction document in [conda-forge overview](https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md#conda-forge-overview).
 
-All the commands in this README should be executed in a terminal with the activated environment.
+### robotology-superbuild (advanced)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,93 +14,23 @@
...
-git clone https://github.dev/robotology/gz-sim-yarp-plugins
-mkdir build
-cd build
-cmake -GNinja -DCMAKE_INSTALL_PREFIX=~/gsyp_ws/install -DCMAKE_PREFIX_PATH=~/gsyp_ws/install ..
-ninja
-ninja install
-~~~
+If you want to build `gz-sim-yarp-plugins` directly from source, you can check the documentation in [`docs/build-from-source.md`](docs/build-from-source.md).
 
 ## Usage
```
traversaro, line 35: The usage part needs to be updated. Specifying the `GZ_SIM_SYSTEM_PLUGIN_PATH` variable manually is only necessary when building from source, in conda and robotology-superbuild this is already handled automatically, I think we need to move this part also in the build from source docs.

docs/build-from-source.md
```diff
@@ -0,0 +1,100 @@
...
+`gz-sim-yarp-plugins` is a fairly classical C++ project build with CMake, so it should be quite easy to build if you are already familiar with how you build C++ projects with CMake.
+If you are not familiar with the use of CMake, you can check some documentation on <https://cmake.org/runningcmake/> or <https://cgold.readthedocs.io> .
+
+## C++ Dependencies
+
+Before building `gz-sim-yarp-plugins`, you need to install its dependencies, the main ones being:
+
+- [Modern Gazebo](https://gazebosim.org/home)
+- [YARP](https://yarp.it/latest//)
+- [YCM-CMake](https://robotology.github.io/ycm/gh-pages/latest/index.html#)
```
traversaro, line 12:
```suggestion
- [YCM CMake Modules](https://robotology.github.io/ycm/gh-pages/latest/index.html#)
```

README.md
xela-95: Oh yes you're right!

xela-95: Done in https://github.com/robotology/gz-sim-yarp-plugins/pull/157/commits/11a014434a1d53f112e333b5c0cf355e07704391

traversaro (approved):

xela-95:
Merging 🚀

PR #177, opened 2024-08-10
 Remove useless checkNetwork calls in plugins that do not require the YARP network

traversaro:
`yarp::os::Network::checkNetwork()` check that the yarpserver (the name server related to YARP ports) is available. However, since in gazebo-yarp-plugins we decouple the regular plugin and the YARP network wrapper server, there is no strict requirement anymore for the yarpserver is up, a user could simply run some devices without any yarpserver.

I noticed this as in my machine the `ConfigurationParsingTest` were segfaulting, until I fixed this. Hopefully this was also the reason why the tests were failing in the macOS CI.


xela-95 (approved):
Thanks a lot @traversaro ! LGTM

xela-95:
Rebased on main. Merging 🚀

PR #178, opened 2024-08-10
Add pixi configuration files

traversaro:
I am probably becoming a bit spoiled, but starting to work on a project and having to read the README to understand how to create a suitable environment instead of just cloning and calling `pixi run test` is becoming a bit annoying for me. So also here I created the required pixi configuration files, opening this draft PR to avoid loosing them.

~This is draft as it also contains https://github.com/robotology/gz-sim-yarp-plugins/pull/177 that is required otherwise the test do not pass.~

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# GitHub syntax highlighting
+pixi.lock linguist-language=YAML linguist-generated=true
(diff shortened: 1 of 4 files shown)


traversaro:
@xela-95 rebase on main, this is ready for review, let me know if you prefer I also document how to use the pixi build (for the CI I would like to first fix the test failures).

xela-95:
Rebased and merged 🚀

PR #184, opened 2024-08-12
implement getEncoderAccelerations

LoreMoretti:


traversaro:
I am not a super big fan of this, if we do not have meaningful accelerometer values having `getEncoderAcceleration` return `false` seems the correct thing to do, why do you need accelerations?

LoreMoretti:
> I am not a super big fan of this, if we do not have meaningful accelerometer values having getEncoderAcceleration return false seems the correct thing to do, why do you need accelerations?

I reported [here](https://github.com/ami-iit/element_torque-control-for-walking/issues/133#issuecomment-2285513564) the issues I am facing. I agree that just setting the accelerations to 0 might not be the best solution. I got inspiration from the gazebo classic yarp plugins: https://github.com/robotology/gazebo-yarp-plugins/blob/d296652b7619dbef39e9bd65ae49f6ca3bdd9377/plugins/controlboard/src/ControlBoardDriverEncoders.cpp#L115-L131

But, I am open to any other solution.

traversaro:
> But, I am open to any other solution.

As neither gazebo-yarp-plugins nor gz-sim-yarp-plugins at the moment provide accelerations, I think it make more sense to disable the use of joint accelerations in wbd by putting `false` in https://github.com/icub-tech-iit/ergocub-software/blob/a49846f5458a3c3fc6bd5feda6704fb7daf64904/urdf/ergoCub/conf/estimators/wbd_ecub_sim.xml#L12 (for now locally, but I think it make sense to open a PR). I think it make a lot of sense to do a PR, as even if the simulators would provide accelerations, for consistency with the real robot it could make sense to set that parameter to false, see https://github.com/robotology/robots-configuration/blob/e5a262e8750102cd36fe64e48e709aeef8e88e29/ergoCubSN002/estimators/wholebodydynamics.xml#L14 .

LoreMoretti:
I agree (and I can open a PR in `ergocub-software`), but I still have one doubt which I have tried to explain [here](https://github.com/ami-iit/element_torque-control-for-walking/issues/133#issuecomment-2285755328).

LoreMoretti:
As @traversaro pointed out it is safer to leave the `getEnconderAccelerations` as it is.

Closing this PR, in favor of https://github.com/icub-tech-iit/ergocub-software/pull/256.

Issue #185, opened 2024-08-12
Close the robotinterface has soon as one of the device attached to it is destroyed


traversaro:
See https://github.com/robotology/gz-sim-yarp-plugins/pull/179#issuecomment-2284529366 .

PR #179, opened 2024-08-10
Add gz-sim-yarp-plugins-check-model executable tool to check if model loads correctly and add CLI11 dependency

traversaro:
This should help ensuring that downstream models using gz-sim-yarp-plugins load correctly, to avoid regression such as https://github.com/icub-tech-iit/ergocub-software/pull/232 .

I still want to iterate a bit on the tool (at the moment it takes in input world files, but I guess it may be also useful to be able to take in input model uri in the form `model://ergoCubGazeboV1` that is the format actually suggest to downstream users to include models in their world `.sdf` files, but the basic logic is already present.

PR #186, opened 2024-08-12
Fix robotinterface plugin crashes by closing it as soon as one of its devices is destroyed

traversaro:
Fix https://github.com/robotology/gz-sim-yarp-plugins/issues/185 .

As the destroying order of the Gazedo entities is not reversed, it occurred that the robotinterface plugin crashed as it was still calling devices that were already destroyed. This PR fixes this problem by closing the robotinterface plugin as soon as one of the devices to which it is attached is destroyed. Basically this is a port of https://github.com/robotology/gazebo-yarp-plugins/pull/619 to gz-sim-yarp-plugins.

### DIFF
--- a/libraries/device-registry/DeviceRegistry.cc
+++ b/libraries/device-registry/DeviceRegistry.cc
@@ -1,9 +1,10 @@
 #include <DeviceRegistry.hh>
 
-#include <cstddef>
 #include <gz/sim/Entity.hh>
 #include <gz/sim/EntityComponentManager.hh>
 #include <gz/sim/Util.hh>
+
+#include <cstddef>
 #include <iostream>
 #include <mutex>
 #include <ostream>
--- a/libraries/device-registry/DeviceRegistry.cc
+++ b/libraries/device-registry/DeviceRegistry.cc
@@ -242,6 +242,7 @@ bool DeviceRegistry::removeDevice(const gz::sim::EntityComponentManager& ecm,
                 return false;
             }
 
+            m_deviceRemovedEvent(deviceDatabaseKey);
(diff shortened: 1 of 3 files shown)


xela-95:
Thanks a lot @traversaro!

In https://github.com/robotology/gazebo-yarp-plugins/pull/619#issuecomment-1086682037 I've seen that to check that the implementation was working as expected, you tried to remove and re-add the model without crashes. Have you tried the same here to check the implementation? We can maybe add some test to spot potential future regressions?

traversaro:
> Thanks a lot @traversaro!
> 
> In the corresponding PR for `gazebo-yarp-plugins` [robotology/gazebo-yarp-plugins#619 (comment)](https://github.com/robotology/gazebo-yarp-plugins/pull/619#issuecomment-1086682037) I've seen that to check that the implementation was working as expected, you tried to remove and re-add the model without crashes. Have you tried the same here to check the implementation? We can maybe add some test to spot potential future regressions?

The problem is that the problem occurs as a result of a race condition between the NWS period threads and gazebo threads. However, the test that we suppressed in https://github.com/robotology/gz-sim-yarp-plugins/pull/179 pass fine with this fix, so that is a sort of test for this.

xela-95:
Just because I've mentioned it before, in gz-sim has been added the equivalent of the Gazebo Classic insert tab (for drag-and-drop of models), it is called `Resource Spawner`, see https://github.com/gazebosim/gz-sim/pull/173#issuecomment-658485569

xela-95 (approved):

xela-95:
Rebased on main.

xela-95:
Merging 🚀

PR #187, opened 2024-08-12
camera: Cleanup memory handling of CameraData

traversaro:
I was starting to look into the Camera plugin to understand a bit of segfaults, and I noticed that the handling of data in `CameraData` was a bit complex. In particular, the only used `CameraData` instance was a local variable of the `gzyarp::Camera` gz-sim plugin, that then is passed via the `setCameraData` function to `gzyarp::CameraDriver` YARP device, where the `m_imageBuffer`  buffer was allocated. To simplify everything, I just moved all the initialization to a `init` method of `CameraData`, and I called `init` in the method of  `gzyarp::Camera` gz-sim plugin.


xela-95 (approved):
Thanks @traversaro ! This simplifies the data management of camera plugin, hopefully reducing the occurrence of segfaults.

xela-95:
Merging 🚀

PR #188, opened 2024-08-12
Add cpp-fmt pixi task to automatically run clang-format

traversaro:
I noticed that I was never running `clang-format` during my PRs, so I added a `cpp-fmt` pixi task (inspired from https://github.com/rerun-io/rerun/blob/354124ee0e667525932d48449339202f27292407/pixi.toml#L379) to automatically run `clang-format`. 

I also run `clang-format` on the existing files, to make sure I did not left anything messy around. Differently from the rest of dependency in the `pixi.toml`, I pinned `clang-format` to the major version 18 (the same version installed by apt in Ubuntu 24.04) as tipically `clang-format` updates can change the formatting.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -42,6 +42,8 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/ca-certificates-2024.7.4-hbcca054_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cairo-1.18.0-hbb29018_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cfitsio-4.4.1-hf8ad068_0.conda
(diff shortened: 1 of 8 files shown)


xela-95:
Thanks @traversaro ! I had in mind to add a github action to automatically check if committed code is compliant to [`.clang-format`](https://github.com/robotology/gz-sim-yarp-plugins/blob/main/.clang-format). 

If I understood correctly, this PR adds a pixi task so that you can format your code locally with clang-format? How do you launch this task? It is using the available format file?

traversaro:
> If I understood correctly, this PR adds a pixi task so that you can format your code locally with clang-format? How do you launch this task? It is using the available format file?

Exactly, this is meant for local use, it can be run `pixi run cpp-fmt`. This just uses clang-format, that I guess should find automatically the `.clang-format` file of the repo.

xela-95 (approved):

xela-95:
Merging 🚀

PR #189, opened 2024-08-13
Fix ClockTest on Windows

traversaro:
Without these changes, the ClockTest is not passing on Windows, the changes are:
* Go back to run the simulation just for ~10 iterations (i.e. 10 ms), and compute correctly the expected nanoseconds published on the clock (without doing `iterations-1` that was what was causing the test to fail on Windows). Furthermore, do not use 1e6 seconds as tolerance for seconds, but just expect them to be equal.
* Add `Finalize()` call to the fixture, otherwise there was a "[Wrn] [D:\bld\gz-sim8_1713968558858\work\src\TestFixture.cc:208] Fixture has not been finalized, any functions you attemptedto hook into will not be run. It is recommended to call Finalize()before accessing the server." warning

### DIFF
--- a/tests/clock/ClockTest.cc
+++ b/tests/clock/ClockTest.cc
@@ -22,14 +22,14 @@ TEST(ClockTest, GetSimulationTimeFromClockPort)
     gz::sim::TestFixture fixture(modelPath.string());
     fixture.Finalize();
 
-    const int iterations = 1000;
+    const int iterations = 10;
     const int deltaTns = 1e6; // 1ms
     const int tolerance = 1e6; // 1ms
     yarp::os::BufferedPort<yarp::os::Bottle> p; // Create a port.
     p.open("/tmp_in"); // Give it a name on the network.
     yarp::os::Network::connect("/clock", "/tmp_in"); // connect two ports.
-    int expectedSimTimeSeconds = (iterations - 1) / 1e3;
-    int expectedSimTimeNanoseconds = 0;
+    int expectedSimTimeSeconds = iterations / 1e3;
+    int expectedSimTimeNanoseconds = static_cast<int>(iterations * deltaTns) % 1'000'000'000;
 
     // ACT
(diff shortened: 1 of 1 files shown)


tests/clock/ClockTest.cc
```diff
@@ -20,15 +20,16 @@ TEST(ClockTest, GetSimulationTimeFromClockPort)
...
+    const int iterations = 10;
     const int deltaTns = 1e6; // 1ms
     const int tolerance = 1e6; // 1ms
     yarp::os::BufferedPort<yarp::os::Bottle> p; // Create a port.
     p.open("/tmp_in"); // Give it a name on the network.
     yarp::os::Network::connect("/clock", "/tmp_in"); // connect two ports.
-    int expectedSimTimeSeconds = (iterations - 1) / 1e3;
-    int expectedSimTimeNanoseconds = 0;
+    int expectedSimTimeSeconds = (iterations) / 1e3;
+    int expectedSimTimeNanoseconds = static_cast<int>(iterations * deltaTns) % 1000000000;
```
xela-95, line 32:
```suggestion
    int expectedSimTimeNanoseconds = static_cast<int>(iterations * deltaTns) % 1e9;
```

xela-95 (approved):
Thanks @traversaro! LGTM! Only minor comments

tests/clock/ClockTest.cc
traversaro: I would prefer to avoid this. `1e9` is a double, not an an `int`. I do not know what happens if you pass a double as second argument of `%`, but I think it is more clear and safe if we use an integere there.

xela-95: Oh yes you're right sorry. I wanted to make more understandable which number is that one. I think that from c++14 it is possible to use single quote to separate digits, hence writing `1'000'000'000`. But this is not blocking for sure.

xela-95:
Rebased and squashed.

xela-95:
Merging 🚀

PR #176, opened 2024-08-08
Align tests and examples to use URIs instead of relative paths

xela-95:
Closes #175

PR #190, opened 2024-08-13
Fix tests on Windows

traversaro:
I was trying to get the tests on Windows to work, and I noticed that the modifications in https://github.com/robotology/gz-sim-yarp-plugins/pull/176/files were unix-specific, as they used `:` as separator for env variables that contain list of paths, while on Windows `;` is used (as `:` could be ambigous with the driver letter separator, i.e. the lists on Windows are `C:\src;C:\otherdir`, not `C:\src:C:\otherdir` that would be ambiguous.

To fix this, I modified the CMake to use ENVIRONMENT_MODIFICATION  and `path_list_append` command, that handles this differences between Windows and *nix automatically, see https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT_MODIFICATION.html . 

To do so, I bumped the minimum CMake version to 3.22 .

Furthermore, I also added `.Finalize()` calls to all the fixtures that were missing them, to get rid of the `"[Wrn] [D:\bld\gz-sim8_1713968558858\work\src\TestFixture.cc:208] Fixture has not been finalized, any functions you attemptedto hook into will not be run. It is recommended to call Finalize()before accessing the server.` warnings.


traversaro:
(there is a spurious commit, please squash and merge).

traversaro:
Even with this fix, the test still segfault on Windows, but I will investigate this once all the PR of fixes are merged.

xela-95:
Thanks a lot @traversaro !! Do you think we should then remove also the the suggestions in the tutorial READMEs, like https://github.com/robotology/gz-sim-yarp-plugins/blob/0691f50548848e96a2558ed2f10f344091f99a5d/tutorial/camera/model/README.md#L38-L46 ?

traversaro:
Yes, I think it make sense to remove also the one in the README, thanks for catching that!

xela-95:
Rebased and squashed the fixups.

xela-95 (approved):
Super @traversaro !! Approving

xela-95:
Merging 🚀

PR #191, opened 2024-08-14
Enable `gz-sim-yarp-plugins-check-model` test on apt CI

xela-95:
This PR enables the  `gz-sim-yarp-plugins-check-model` test introduced in #179 and excluded from apt CI for an issue solved in #185.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -108,9 +108,7 @@ jobs:
       run: |
         cd build
         # So high until-pass as a mitigation for https://github.com/robotology/gz-sim-yarp-plugins/issues/75
-        # For some reason gz-sim-yarp-plugins-check-model-tutorial-forcetorque-model-two-sensors fails only in apt CI in Debug,
-        # let's just skip it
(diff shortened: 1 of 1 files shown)


xela-95:
Now it seems that this test is now passing also on apt.

traversaro (approved):

traversaro:
I suspect we could try (even in another PR) also to reduce the ` --repeat until-pass:` settings, as many flaky failure I think could be due to the race condition fixed in https://github.com/robotology/gz-sim-yarp-plugins/pull/186 .

xela-95:
> I suspect we could try (even in another PR) also to reduce the ` --repeat until-pass:` settings, as many flaky failure I think could be due to the race condition fixed in #186 .

do you mean to remove the `--repeat until-pass` arguments in test for both apt CI and conda CI?

traversaro:
> > I suspect we could try (even in another PR) also to reduce the ` --repeat until-pass:` settings, as many flaky failure I think could be due to the race condition fixed in #186 .
> 
> do you mean to remove the `--repeat until-pass` arguments in test for both apt CI and conda CI?

It is something that we could try, but it is independent of this PR.

xela-95:
Merging! 🚀

Issue #182, opened 2024-08-12
issue when running gz-sim after exporting YARP_CLOCK environment variable


LoreMoretti:
I started `yarpserver` and run `gz sim ergocubGazevoV1_1_gz_sim_world.sdf`, what I get is the following:

![image](https://github.com/user-attachments/assets/9a273ff5-c298-4ed9-b9ed-8e87770ddd59)

The gz-sim window remains black, and the following gets printed:

```
[INFO] |yarp.os.Port|/IITICUBLAP226/gz/62301/clock:i| Port /IITICUBLAP226/gz/62301/clock:i active at tcp://172.20.10.5:10002/
[INFO] |yarp.os.Network| Success: port-to-port persistent connection added.
[ERROR] |yarp.os.NetworkClock| Cannot find time port "/clock" or a time topic "/clock@"
[INFO] |yarp.os.Time| Waiting for clock server to start broadcasting data ...
[INFO] |yarp.os.Time| Waiting for clock server to start broadcasting data ...
```

with this line `[INFO] |yarp.os.Time| Waiting for clock server to start broadcasting data ...` being repeatedly printed on the terminal.

traversaro:
I guess this is the same issue of https://github.com/robotology/gazebo-yarp-plugins/issues/526 that was fixed for Gazebo Classic in https://github.com/robotology/gazebo-yarp-plugins/pull/537, probably we can do a similar fix and a test similar to https://github.com/robotology/gazebo-yarp-plugins/pull/537 ?

PR #192, opened 2024-08-14
Avoid clock plugin deadlock when YARP_CLOCK is already set

xela-95:
Fix https://github.com/robotology/gz-sim-yarp-plugins/issues/182.


xela-95:
Up to now I've checked that by commenting out line https://github.com/robotology/gz-sim-yarp-plugins/pull/192/files#diff-e544b1b964ccc8b1b9805630527cfbd807c46fa4333c455e5e60d185304578e3R57 
```cpp
    yarp::conf::environment::set_string("YARP_CLOCK", "/clock");
```
in the new regression test, the test passes.

Now I have to find how to correctly handle this.

tests/clock/ClockTest.cc
```diff
@@ -43,3 +44,35 @@ TEST(ClockTest, GetSimulationTimeFromClockPort)
...
+    auto modelPath = std::filesystem::path(CMAKE_CURRENT_SOURCE_DIR) / "model.sdf";
+    gz::sim::TestFixture fixture(modelPath.string());
+    fixture.Finalize();
+
+    const int iterations = 10;
+    const int deltaTns = 1e6; // 1ms
+
+    // ACT
+    fixture.Server()->Run(/*_blocking=*/true, iterations, /*_paused=*/false);
+    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
traversaro, line 72: Probably we can try to reduce this? The test are already spending a lot of time, and in this specific case I do not think we need to wait for anything (the port registration is blocking as far as I know).

xela-95:
Up to now the test if generating an [exception](https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/10386139607/job/28756766013?pr=192#step:12:111):

```console
[Dbg] [Sensors.cc:262] Waiting for init
[Dbg] [SystemManager.cc:74] Loaded system [gz::sim::systems::SceneBroadcaster] for entity [1]
[INFO] |yarp.os.Port|/fv-az693-479//home/runner/work/gz-sim-yarp-plugins/gz-sim-yarp-plugins/build/bin/ClockTest/12999/clock:i| Port /fv-az693-479//home/runner/work/gz-sim-yarp-plugins/gz-sim-yarp-plugins/build/bin/ClockTest/12999/clock:i active at tcp://127.0.0.1:10002/
Error:  |yarp.os.Network| cannot find port /root
Failure: name server did not accept connection to topic.
Error:  |yarp.os.NetworkClock| Cannot find time port "/clock" or a time topic "/clock@"
[FATAL] |yarp.os.Time| failed creating NetworkClock client, cannot open input port

```

tests/clock/ClockTest.cc
traversaro: As an alternative, we can check for the port existence in a for and just wait a small amount of time for each check, so that as soon as that port appears the test exits, and we need to wait for the full time only if the test fails.

xela-95: ok! I'll try to see if it works without delay, and in case it is necessary, I will implement your solution.

traversaro:
Perhaps this the reason why I used an external yarpserver in the gazebo-classic test?

xela-95:
> Perhaps this the reason why I used an external yarpserver in the gazebo-classic test?

Yep it could be 😅

xela-95:
In https://github.com/robotology/gz-sim-yarp-plugins/pull/192/commits/21dba7bb132447ad771f3f755a5a020323f7ffa5 I tried to launch `yarp server` from a separate process using [tiny-process-library](https://gitlab.com/eidheim/tiny-process-library), repeating what has been done in gazebo-yarp-plugins. Unfortunately this does not fix the exception I was getting before:

```
13: [INFO] |yarp.os.Port|/IITICUBLAP218//home/acroci/repos/gz-sim-yarp-plugins/build/bin/ClockTest/586888/clock:i| Port /IITICUBLAP218//home/acroci/repos/gz-sim-yarp-plugins/build/bin/ClockTest/586888/clock:i active at tcp://127.0.0.1:10002/
13: [ERROR] |yarp.os.Network| cannot find port /root
13: Failure: name server did not accept connection to topic.
13: [ERROR] |yarp.os.NetworkClock| Cannot find time port "/clock" or a time topic "/clock@"
13: [FATAL] |yarp.os.Time| failed creating NetworkClock client, cannot open input port
```

xela-95:
I'm not fully convinced of the return value of [yarp::os::NetworkBase::isNetworkInitialized()](https://github.com/robotology/yarp/blob/5d8412864a8980f08211cfd1f3431fd32f586441/src/libYARP_os/src/yarp/os/Network.cpp#L876C35-L879), since it is always returning true, even if I do not start a `yarp server`.

xela-95:
Thasnk to @traversaro suggestions I've been able to adapt the clock plugin to work without deadlocks also in case of the YARP_CLOCK variable already set. Ready for review.

traversaro (approved):

xela-95:
Merging 🚀

Issue #71, opened 2024-01-25
Modify sensor plugins to avoid use of non-deterministic functionality to access the data


traversaro:
We are experiencing many flaky errors in gz-sim-yarp-plugins tests:
* https://github.com/robotology/gz-sim-yarp-plugins/issues/54
* https://github.com/robotology/gz-sim-yarp-plugins/issues/31
* https://github.com/robotology/gz-sim-yarp-plugins/issues/42
* https://github.com/robotology/gz-sim-yarp-plugins/issues/41

To be honest, I did not looked a lot into those back in time when I opened the issue. However, as @xela-95 started working in the repo, I noticed that most sensor plugins are actually using the gz's gz-transport (i.e. ZeroMQ + ProtoBuf) functionality to actually access sensors. To give an example, let's look at the forcetorque plugin. 
Before the first step of simulation, a connection via gz-transport is established in the main physics thread of gz-sim by the following snippet of code (from https://github.com/robotology/gz-sim-yarp-plugins/blob/25216f1c90ed524a30b664e7cc327b368eb2d649/plugins/forcetorque/ForceTorque.cc#L139):
~~~
        if (!this->ftInitialized && _ecm.ComponentData<components::SensorTopic>(sensor).has_value())
        {
            this->ftInitialized = true;
            auto ftTopicName = _ecm.ComponentData<components::SensorTopic>(sensor).value();
            this->node.Subscribe(ftTopicName, &ForceTorque::ftCb, this);
        }
~~~

This will call the `ForceTorque::ftCb` callback once a message is received, that will actually just do:
~~~
    void ftCb(const gz::msgs::Wrench& _msg)
    {
        std::lock_guard<std::mutex> lock(this->ftMsgMutex);
        ftMsg = _msg;
    }
~~~

However, this will happen (from what I understand) in a thread **different** from the main physics thread.

Then, after the update of the physics, the following code again in the **main physics thread** reads the ft values (that are just ~6 doubles) from the buffer that was populated by `ftCb` :
~~~
        gz::msgs::Wrench ftMsg;
        {
            std::lock_guard<std::mutex> lock(this->ftMsgMutex);
            ftMsg = this->ftMsg;
        }

        std::lock_guard<std::mutex> lock(forceTorqueData.m_mutex);
        forceTorqueData.m_data[0] = (ftMsg.force().x() != 0) ? ftMsg.force().x() : 0;
[...]

PR #193, opened 2024-08-20
Properly initialize to false ftInitialized variable

LoreMoretti:
This fixes the wrong stream of FT data

### DIFF
--- a/plugins/forcetorque/ForceTorque.cc
+++ b/plugins/forcetorque/ForceTorque.cc
@@ -196,7 +196,7 @@ private:
     std::string sensorScopedName;
     bool m_deviceRegistered;
     ForceTorqueData forceTorqueData;
-    bool ftInitialized;
+    bool ftInitialized{false};
     gz::transport::Node node;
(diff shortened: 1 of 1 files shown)


traversaro:
Just to clarify, without this fix it could happen that a FT sensor would start and always return a valid measurement of all 0, if the `ftInitialized` variable actually contained `true`. Unfortunately, we never realized this as any kind of non-deterministic was due to https://github.com/robotology/gz-sim-yarp-plugins/issues/71 and we added workaround in the tests. 

To avoid this kind of problems in the future, we could:
* Make sure all primitive types are initialized
* Run all tests under valgrind, as we do in other projects (idyntree and blf)
* Remove all sources of non-determinism, so we can avoid using `--until-pass` in running the tests, that hide this kind of ailures

Anyhow, this are all things we can do in the future. In the meanwhile, as this is quite a critical bug, I would merge directly as @xela-95 is current out of office.

traversaro (approved):

Issue #54, opened 2023-12-21
macOS CI fails with ImuTest error


traversaro:
macOS CI fails with error:
~~~
[INFO] |yarp.os.impl.PortCoreInputUnit|/IMU/rpc:o| Receiving input from /ImuTest/rpc:i to /IMU/rpc:o using tcp
[INFO] |yarp.os.impl.PortCoreOutputUnit|/ImuTest/rpc:i| Sending output from /ImuTest/rpc:i to /IMU/rpc:o using tcp
[INFO] |yarp.os.impl.PortCoreInputUnit|/ImuTest/measures:i| Receiving input from /IMU/measures:o to /ImuTest/measures:i using tcp
[INFO] |yarp.os.impl.PortCoreOutputUnit|/IMU/measures:o| Sending output from /IMU/measures:o to /ImuTest/measures:i using tcp
[DEBUG] |yarp.device.multipleanalogsensorsclient| Open complete
[INFO] |yarp.dev.PolyDriver|multipleanalogsensorsclient| Created device <multipleanalogsensorsclient>. See C++ class MultipleAnalogSensorsClient for documentation.
[Wrn] [TestFixture.cc:215] Fixture has not been finalized, any functions you attemptedto hook into will not be run. It is recommended to call Finalize()before accessing the server.
/Users/runner/work/gz-sim-yarp-plugins/gz-sim-yarp-plugins/tests/imu/ImuTest.cc:77: Failure
The difference between measureOrientation(2) and 1.0 is 1, which exceeds 1e-2, where
measureOrientation(2) evaluates to 0,
1.0 evaluates to 1, and
1e-2 evaluates to 0.01.
/Users/runner/work/gz-sim-yarp-plugins/gz-sim-yarp-plugins/tests/imu/ImuTest.cc:82: Failure
The difference between measureAccelerometer(2) and 9.8 is 9.8000000000000007, which exceeds 1e-2, where
measureAccelerometer(2) evaluates to 0,
9.8 evaluates to 9.8000000000000007, and
1e-2 evaluates to 0.01.
[INFO] |yarp.os.impl.PortCoreInputUnit|/ImuTest/measures:i| Removing input from /IMU/measures:o to /ImuTest/measures:i
[INFO] |yarp.os.impl.PortCoreOutputUnit|/IMU/measures:o| Removing output from /IMU/measures:o to /ImuTest/measures:i
[...]

PR #195, opened 2024-08-20
Reduce test skip and options to tolerate non-deterministic failure

traversaro:
We recently have a few PRs that solved non-deterministic failures:
* https://github.com/robotology/gz-sim-yarp-plugins/pull/193
* https://github.com/robotology/gz-sim-yarp-plugins/pull/194
* https://github.com/robotology/gz-sim-yarp-plugins/pull/186

Thanks to these fixes, we can reduce the number of workarounds, and only run multiple times tests that are known to be flaky.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -100,15 +100,13 @@ jobs:
       if: contains(matrix.build_type, 'Release')
       run: |
         cd build
-        # So high until-pass as a mitigation for https://github.com/robotology/gz-sim-yarp-plugins/issues/75
-        ctest -E "^ImuTest" --repeat until-pass:20 --output-on-failure -C ${{ matrix.build_type }} .
+        ctest --output-on-failure -C ${{ matrix.build_type }} .
 
     - name: Test (Debug)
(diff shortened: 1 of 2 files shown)


traversaro:
Ok, apparently `CameraTest` still segfaults on apt, while `ConcurrentInstancesTest` fails like:

~~~
ODE INTERNAL ERROR 1: assertion "!colliders_initialized" failed in dInitColliders() [collision_kernel.cpp:168]
~~~

that is due to https://github.com/gazebosim/gz-sim/issues/18 .

traversaro:
Ok, the good news is that all test pass on macos now, so https://github.com/robotology/gz-sim-yarp-plugins/issues/54 is fixed. On conda we still have problems with LaserTest and CameraTest, while on apt with `CameraTest` and `ControlBoardOnMultipleGazeboInstancesTest`/`ConcurrentInstancesTest`.

traversaro:
Ok, now the CI passes, and in the tests we just skip/permit failure only for tests for which we have open issue tracking the problems.

LoreMoretti (approved):

Issue #201, opened 2024-08-21
ConcurrentInstancesTest and ControlBoardOnMultipleGazeboInstancesTest non-deterministic crash


traversaro:
Similar to https://github.com/robotology/gz-sim-yarp-plugins/issues/200, these two tests still crash non deterministically. Unfortunately, the problem is upstream in gz-physics/dartsim/ode (see https://github.com/gazebosim/gz-sim/issues/18 and https://github.com/robotology/gz-sim-yarp-plugins/pull/153) .

While this does not fix the problem in general, I think we should at least switch to another DART's collision detector (not ODE, that has this multithreading problems) in the tests, to ensure that they run reliably.

traversaro:
Apparently the world already have some tags:

~~~xml
        <physics type="ignored">
            <dart>
                <collision_detector>dart</collision_detector>
            </dart>
        </physics>
~~~

but as the tests are still failing with ode-related errors, probably they are not read correctly.

traversaro:
> Apparently the world already have some tags:
> 
> ```
>         <physics type="ignored">
>             <dart>
>                 <collision_detector>dart</collision_detector>
>             </dart>
>         </physics>
> ```
> 
> but as the tests are still failing with ode-related errors, probably they are not read correctly.

The reason is described (and fixed) in https://github.com/gazebosim/gz-physics/pull/675 . However, even if the PR is accepted it will take some time to get release, for the time being we can continue with `--repeat until-success:20`.

PR #202, opened 2024-08-23
Add workaround for "ODE INTERNAL ERROR 1: assertion "!colliders_initialized" failed in dInitColliders() [collision_kernel.cpp:168]" crash

traversaro:
This PR adds a simple workaround for the non-deterministic crashes with messages:

> ODE INTERNAL ERROR 1: assertion "!colliders_initialized" failed in dInitColliders() [collision_kernel.cpp:168]

described in https://github.com/robotology/gz-sim-yarp-plugins/issues/201 . The upstream solution is https://github.com/gazebosim/gz-physics/pull/675, but while we wait for upstream feedback, this simple solution of running the first step as blocking so the two first step happen in two different times for sure should work fine. 

Fix https://github.com/robotology/gz-sim-yarp-plugins/issues/201 .

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -101,18 +101,18 @@ jobs:
       run: |
         cd build
         # Deterministic tests
-        ctest -E "^ConcurrentInstancesTest|^ControlBoardOnMultipleGazeboInstancesTest|^CameraTest" --output-on-failure -C ${{ matrix.build_type }} .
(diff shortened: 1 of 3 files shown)


traversaro:
Great, I tried a bit locally and the workaround is working.

xela-95 (approved):
Thanks a lot @traversaro !! I also tested this PR locally on conda and pixi and the tests passed!

xela-95:
Merging 🚀

Issue #196, opened 2024-08-20
LaserTest and CameraTest are crashing in CI when using conda-forge dependencies on Linux


traversaro:

traversaro:
On WSL, the test are crashing on exit if `LIBGL_ALWAYS_SOFTWARE=true` is not set, probably something related to d3d12 or the intel or nvidia driver (as forcing the use of nvidia GPU via `export MESA_D3D12_DEFAULT_ADAPTER_NAME=NVIDIA` does not change anything).

Instead, on CI the test are crashing with:

~~~
 [Wrn] [Ogre2RenderEngine.cc:551] Unable to open display: . Trying to run in headless mode.
[Err] [Ogre2RenderEngine.cc:752] Unable to load Ogre Plugin[/home/runner/micromamba/envs/gzyarppluginsdev/lib/OGRE-Next/RenderSystem_GL3Plus]. Rendering will not be possible.Make sure you have installed OGRE properly.
[Err] [Ogre2RenderEngine.cc:803] unable to find OpenGL 3+ Rendering Subsystem. OGRE is probably installed incorrectly. Double check the OGRE cmake output, and make sure OpenGL is enabled.
~~~

I have no idea why we do not have the same problem in apt.

traversaro:
The related code i https://github.com/gazebosim/gz-rendering/blob/64820c0f3b389f32ec929130c8486c42df43e9e5/ogre2/src/Ogre2RenderEngine.cc#L550-L552 .

traversaro:
Given that the error is `[Err] [Ogre2RenderEngine.cc:752] Unable to load Ogre Plugin[/home/runner/micromamba/envs/gzyarppluginsdev/lib/OGRE-Next/RenderSystem_GL3Plus]. Rendering will not be possible.Make sure you have installed OGRE properly.`, I added the `ldd $CONDA_PREFIX/lib/OGRE-Next/RenderSystem_GL3Plus.so` in CI to understand if there is some library missing.

traversaro:
[...]

Issue #200, opened 2024-08-21
LaserTest and CameraTest non deterministic crash


traversaro:
After https://github.com/robotology/gz-sim-yarp-plugins/pull/195, these two tests still fail non deterministically. This issue track fixing this flaky failures.

traversaro:
The problem is still there:

~~~
 4/18 Test  #4: CameraTest ................................................................***Exception: SegFault  2.39 sec
Running main() from ./googletest/src/gtest_main.cc
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from CameraTest
[ RUN      ] CameraTest.PluginTest
[Msg] Loading SDF world file[/home/runner/work/gz-sim-yarp-plugins/gz-sim-yarp-plugins/tests/camera/model.sdf].
[Msg] Serving entity system service on [/entity/system/add]
[Dbg] [Physics.cc:870] Loaded [gz::physics::dartsim::Plugin] from library [/usr/lib/x86_64-linux-gnu/gz-physics-7/engine-plugins/libgz-physics-dartsim-plugin.so]
[Dbg] [SystemManager.cc:77] Loaded system [gz::sim::systems::Physics] for entity [1]
[Msg] Create service on [/world/sensors/create]
[Msg] Remove service on [/world/sensors/remove]
[Msg] Pose service on [/world/sensors/set_pose]
[Msg] Pose service on [/world/sensors/set_pose_vector]
[Msg] Light configuration service on [/world/sensors/light_config]
[Msg] Physics service on [/world/sensors/set_physics]
[Msg] SphericalCoordinates service on [/world/sensors/set_spherical_coordinates]
[Msg] Enable collision service on [/world/sensors/enable_collision]
[Msg] Disable collision service on [/world/sensors/disable_collision]
[Msg] Material service on [/world/sensors/visual_config]
[Msg] Material service on [/world/sensors/wheel_slip]
[Dbg] [SystemManager.cc:77] Loaded system [gz::sim::systems::UserCommands] for entity [1]
[Dbg] [Sensors.cc:530] Configuring Sensors system
[Dbg] [Sensors.cc:449] SensorsPrivate::Run
[Dbg] [SystemManager.cc:77] Loaded system [gz::sim::systems::Sensors] for entity [1]
[Dbg] [Sensors.cc:425] SensorsPrivate::RenderThread started
[Dbg] [Sensors.cc:265] Waiting for init
[...]

PR #203, opened 2024-08-23
Run CameraTest and LaserTest as normal tests

traversaro:
After https://github.com/robotology/gz-sim-yarp-plugins/issues/196, let's check if there still remain problems in CameraTest or LaserTest.


traversaro:
Ok, `CameraTest` still segfaults sometimes on apt, this is tracked in https://github.com/robotology/gz-sim-yarp-plugins/issues/200 .

PR #207, opened 2024-09-03
Add github workflow to automatically update pixi lockfile

xela-95:
In this way, we do not risk forgetting to maintain the Pixi lock files.

The action comes from the official documentation: https://pixi.sh/dev/advanced/updates_github_actions/#how-to-use with some customizations done by @flferretti in https://github.com/ami-iit/jaxsim/pull/221

Remark: as specified in the documentation, the permission of the repo should be set as follows:
image

### DIFF
--- a/.github/workflows/update-pixi-lockfile.yml
+++ b/.github/workflows/update-pixi-lockfile.yml
@@ -1,0 +1,42 @@
+name: Update lockfiles
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 1 * *
+
+jobs:
+  pixi-update:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up pixi
+        uses: prefix-dev/setup-pixi@v0.8.1
+        with:
+          run-install: false
+
+      - name: Install pixi-diff-to-markdown
+        run: pixi global install pixi-diff-to-markdown
+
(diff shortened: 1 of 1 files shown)


xela-95:
Triggered the action on my fork of the repository since the action is only available when on main: https://github.com/xela-95/gz-sim-yarp-plugins/actions/runs/10724133024

xela-95:
> Triggered the action on my fork of the repository since the action is only available when on main: https://github.com/xela-95/gz-sim-yarp-plugins/actions/runs/10724133024

It executed successfully and it generated the PR: https://github.com/xela-95/gz-sim-yarp-plugins/pull/1

traversaro (approved):

traversaro:
I am not enthusiastic of the idea of the automatic pixi-lock update (for the reasons described in https://github.com/robotology/whole-body-estimators/pull/191), but if it is happening only one a month that is ok for me.

xela-95:
> I am not enthusiastic of the idea of the automatic pixi-lock update (for the reasons described in [robotology/whole-body-estimators#191](https://github.com/robotology/whole-body-estimators/pull/191)), but if it is happening only one a month that is ok for me.

Thanks for linking the discussion, I've not considered that side-effect, maybe we could put track the pixi lockfile with LFS or maybe only maintain the the manual worflow-dispatch way of running the action.

traversaro:
I think we can try the automatic update for some time and then monitor the size of the repo. If it increases too much, we can switch to manual sporadic updates.

xela-95:
Perfect! Merging 🚀

PR #221, opened 2024-12-02
ControlBoardDriver: add iJointCoupling interface

Nicogene:
This PR adds the [`iJointCoupling` interface](https://github.com/robotology/yarp/blob/master/src/libYARP_dev/src/yarp/dev/IJointCoupling.h), that allows to handle coupled systems using YARP device driver such as:
- https://github.com/icub-tech-iit/ergocub-software/blob/master/src/modules/couplingXCubHandMk5/README.md

### DIFF
--- a/plugins/controlboard/include/ControlBoardData.hh
+++ b/plugins/controlboard/include/ControlBoardData.hh
@@ -62,7 +62,7 @@ struct PhysicalJointProperties
     std::string positionControlLaw; // TODO: verify usefulness of this field
 };
 
-struct ActuatedAxesProperties
+struct ActuatedAxisProperties
 {
     CommonJointProperties commonJointProperties;
     std::unique_ptr<yarp::dev::gzyarp::TrajectoryGenerator> trajectoryGenerator;
@@ -77,14 +77,17 @@ class ControlBoardData
 public:
     std::mutex mutex;
     std::vector<PhysicalJointProperties> physicalJoints;
-    std::vector<ActuatedAxesProperties>  actuatedAxes;
+    std::vector<ActuatedAxisProperties>  actuatedAxes;
     yarp::os::Stamp simTime;
     yarp::dev::IJointCoupling* ijointcoupling{nullptr};
     // TODO (xela95): read this value from configuration file
     std::chrono::milliseconds controlUpdatePeriod = std::chrono::milliseconds(1);
-
+    
+    bool initCoupledJoints();
     bool setInteractionMode(int axis, yarp::dev::InteractionModeEnum mode);
     bool setControlMode(int j, int mode);
+private:
+    yarp::sig::VectorOf<size_t> coupledActuatedAxes, coupledPhysicalJoints;
 };
 
 class IControlBoardData
--- a/plugins/controlboard/include/ControlBoardData.hh
+++ b/plugins/controlboard/include/ControlBoardData.hh
@@ -34,8 +34,7 @@ struct PidControlTypeEnumHashFunction
     }
 };
 
-struct JointProperties
-{
+struct CommonJointProperties {
     std::string name;
     yarp::dev::InteractionModeEnum interactionMode{yarp::dev::InteractionModeEnum::VOCAB_IM_STIFF};
     yarp::conf::vocab32_t controlMode{VOCAB_CM_IDLE};
@@ -53,9 +52,19 @@ struct JointProperties
     double velocity{0.0}; // Joint velocity [deg/s]
     double velocityLimitMin{std::numeric_limits<double>::min()};
     double velocityLimitMax{std::numeric_limits<double>::max()};
+}
+
+struct PhysicalJointProperties
+{
+    CommonJointProperties commonJointProperties;
     std::unordered_map<yarp::dev::PidControlTypeEnum, gz::math::PID, PidControlTypeEnumHashFunction>
         pidControllers;
     std::string positionControlLaw; // TODO: verify usefulness of this field
+};
+
+struct ActuatedAxesProperties
+{
+    CommonJointProperties commonJointProperties;
     std::unique_ptr<yarp::dev::gzyarp::TrajectoryGenerator> trajectoryGenerator;
     double trajectoryGenerationRefPosition{0.0};
     double trajectoryGenerationRefSpeed{0.0};
@@ -67,11 +76,10 @@ class ControlBoardData
 {
 public:
     std::mutex mutex;
-    std::vector<JointProperties> physicalJoints;
-    std::vector<JointProperties> actuatedAxes;
+    std::vector<PhysicalJointProperties> physicalJoints;
+    std::vector<ActuatedAxesProperties>  actuatedAxes;
     yarp::os::Stamp simTime;
     yarp::dev::IJointCoupling* ijointcoupling{nullptr};
-    std::vector<JointProperties>* actuated_joints_handle;
     // TODO (xela95): read this value from configuration file
     std::chrono::milliseconds controlUpdatePeriod = std::chrono::milliseconds(1);
 };
--- a/plugins/controlboard/include/ControlBoardData.hh
+++ b/plugins/controlboard/include/ControlBoardData.hh
@@ -82,6 +82,9 @@ public:
     yarp::dev::IJointCoupling* ijointcoupling{nullptr};
     // TODO (xela95): read this value from configuration file
     std::chrono::milliseconds controlUpdatePeriod = std::chrono::milliseconds(1);
+
+    bool setInteractionMode((int axis, yarp::dev::InteractionModeEnum mode);
+    bool setControlMode(int j, int mode);
 };
 
 class IControlBoardData
--- a/plugins/controlboard/include/ControlBoardData.hh
+++ b/plugins/controlboard/include/ControlBoardData.hh
@@ -71,6 +71,7 @@ public:
     std::vector<JointProperties> actuatedAxes;
     yarp::os::Stamp simTime;
     yarp::dev::IJointCoupling* ijointcoupling{nullptr};
+    std::vector<JointProperties>* actuated_joints_handle;
     // TODO (xela95): read this value from configuration file
     std::chrono::milliseconds controlUpdatePeriod = std::chrono::milliseconds(1);
 };
--- a/plugins/controlboard/include/ControlBoardData.hh
+++ b/plugins/controlboard/include/ControlBoardData.hh
@@ -52,7 +52,7 @@ struct CommonJointProperties {
     double velocity{0.0}; // Joint velocity [deg/s]
     double velocityLimitMin{std::numeric_limits<double>::min()};
     double velocityLimitMax{std::numeric_limits<double>::max()};
-}
+};
 
 struct PhysicalJointProperties
 {
@@ -83,7 +83,7 @@ public:
     // TODO (xela95): read this value from configuration file
     std::chrono::milliseconds controlUpdatePeriod = std::chrono::milliseconds(1);
 
-    bool setInteractionMode((int axis, yarp::dev::InteractionModeEnum mode);
+    bool setInteractionMode(int axis, yarp::dev::InteractionModeEnum mode);
     bool setControlMode(int j, int mode);
 };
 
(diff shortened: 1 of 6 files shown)


Nicogene:
The latest commit seems a reasonable implementation, it is all base on this line of code (that I hope it works as I think):
```c++
    m_controlBoardData.actuated_joints_handle = m_controlBoardData.ijointcoupling ? &(m_controlBoardData.actuatedAxes) : &(m_controlBoardData.physicalJoints); 
```

Because I introduced a second set of joints (physicalJoints and actuatedAxes) and in function of the presence or not of coupling they are the same thing or not.

For now I didn't introduced the decoupling of the velocities because it seems not used

Nicogene:
Yesterday @traversaro @xela-95 and I had an alignment, and we decided to deeply refactor the plugin in order to make cleaner the introduction of the coupling.

I managed to have a first version that compiles, it is already a success given how many things I changed :D 

I will now test it using the current configuration files of ergocub (no coupling involved), in case of positive outcome I will move to test it with the coupling of the hands mk5.x.

Nicogene:
> If you want a more self-contained test with a smaller model w.r.t. to ergocub, it seems that also the single_pendulum tests is currently failing.

Thanks for the heads up! I used that test for debugging it and I managed to find that specific error, one controlboard test is still failing I am not sure if it is due to my changes

Nicogene:
I managed to fix:
- limits
- names

for the actuated axes, I am still getting a segmentation fault on `ControlBoardCommonsTest`

At first glance, ergocub seems controllable with my changes
image

Nicogene:
The coupling seems to work just fine

image

Nicogene:
I noticed that enabling the control of the hands make drop the RTF of a ~ 20%, this can be noticed also in the screenshots I posted:
- [No hands](https://github.com/robotology/gz-sim-yarp-plugins/pull/221#issuecomment-2541752185) RTF **79%**
- [With hands](https://github.com/robotology/gz-sim-yarp-plugins/pull/221#issuecomment-2545211545) RTF **58%**

I have also checked the RTF on `main` and it is ~ 78% so it seems not that my changes slowed down the simulation, probably the more joints to control make the simulation slower


cc @xela-95 @traversaro

Nicogene:
> Great! Not blocking, but most of the (implicit) documentation for plugins is given in the `tutorial` folder in form of examples for each plugin. To have an example for the coupling functionality, we could port [robotology/gazebo-yarp-plugins@`master`/tutorial/model/coupled_pendulum](https://github.com/robotology/gazebo-yarp-plugins/tree/master/tutorial/model/coupled_pendulum) (and obviously it will start depending on icub-main). That will provide also a simple smoke test for the coupling functionality in
> 
> https://github.com/robotology/gz-sim-yarp-plugins/blob/64deb83de3d4bbb606c14ee6f3605212b1004587/tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt#L19
> .

Ok for me! But we need first to merge the coupling device PR in icub-main:
- https://github.com/robotology/icub-main/pull/997

traversaro:
> > Great! Not blocking, but most of the (implicit) documentation for plugins is given in the `tutorial` folder in form of examples for each plugin. To have an example for the coupling functionality, we could port [robotology/gazebo-yarp-plugins@`master`/tutorial/model/coupled_pendulum](https://github.com/robotology/gazebo-yarp-plugins/tree/master/tutorial/model/coupled_pendulum) (and obviously it will start depending on icub-main). That will provide also a simple smoke test for the coupling functionality in
> > https://github.com/robotology/gz-sim-yarp-plugins/blob/64deb83de3d4bbb606c14ee6f3605212b1004587/tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt#L19
> > 
> > .
> 
> Ok for me! But we need first to merge the coupling device PR in icub-main:
> 
>     * [Add `couplingICubEye` device icub-main#997](https://github.com/robotology/icub-main/pull/997)

Sure! We can also just add tutorial and then add the tests once https://github.com/robotology/icub-main/pull/997 is merged.

Nicogene:
> > > Great! Not blocking, but most of the (implicit) documentation for plugins is given in the `tutorial` folder in form of examples for each plugin. To have an example for the coupling functionality, we could port [robotology/gazebo-yarp-plugins@`master`/tutorial/model/coupled_pendulum](https://github.com/robotology/gazebo-yarp-plugins/tree/master/tutorial/model/coupled_pendulum) (and obviously it will start depending on icub-main). That will provide also a simple smoke test for the coupling functionality in
> > > https://github.com/robotology/gz-sim-yarp-plugins/blob/64deb83de3d4bbb606c14ee6f3605212b1004587/tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt#L19
> > > 
> > > .
> > 
> > 
> > Ok for me! But we need first to merge the coupling device PR in icub-main:
> > ```
> > * [Add `couplingICubEye` device icub-main#997](https://github.com/robotology/icub-main/pull/997)
> > ```
> 
> Sure! We can also just add tutorial and then add the tests once [robotology/icub-main#997](https://github.com/robotology/icub-main/pull/997) is merged.

I would prefer to add the tutorial after merging https://github.com/robotology/icub-main/pull/997 because I am not 100% about the configuration file for that coupling, I can add the tutorial but it is possible that the configuration files has  to be revised after testing the coupling of the eyes

traversaro:
Ok, no problem for me!

plugins/controlboard/include/ControlBoardData.hh
```diff
@@ -52,9 +52,19 @@ struct JointProperties
...
+
+struct PhysicalJointProperties
+{
+    CommonJointProperties commonJointProperties;
     std::unordered_map<yarp::dev::PidControlTypeEnum, gz::math::PID, PidControlTypeEnumHashFunction>
         pidControllers;
     std::string positionControlLaw; // TODO: verify usefulness of this field
+};
+
+struct ActuatedAxesProperties
```
xela-95, line 65: Since this is referred to a single axis maybe we should maintain the singular?

```suggestion
struct ActuatedAxisProperties
```

plugins/controlboard/src/ControlBoard.cpp
```diff
@@ -132,6 +132,25 @@ void ControlBoard::Configure(const Entity& _entity,
         return;
     }
 
+    auto& coupling_group_bottle = m_pluginParameters.findGroup("COUPLING");
+    if (!coupling_group_bottle.isNull()) {
+        m_pluginParameters.unput("device");
```
traversaro, line 137: Can we avoid to change the `m_pluginParameters` attribute? If we need to get all the parameters of `m_pluginParameters`, we can copy this to a `m_couplingParameters` attribute, and then set the device just there. However, just to understand, why do we need to pass the whole plugin parameters to the coupling device?

plugins/controlboard/src/ControlBoard.cpp
```diff
@@ -132,6 +132,25 @@ void ControlBoard::Configure(const Entity& _entity,
...
+        auto coupling_device_str = coupling_group_bottle.find("device").asString();
+        m_pluginParameters.put("device", coupling_device_str);
+        if(!m_coupling_driver.open(m_pluginParameters))
+        {
+            yError() << "gz-sim-yarp-controlboard-system Plugin failed: error in opening yarp "
+                        "coupling driver";
```
traversaro, line 143:
```suggestion
                        "coupling driver with device name " << coupling_device_str 
```

I think this may be useful for debug if something does not work.

plugins/controlboard/src/ControlBoardData.cpp
```diff
@@ -0,0 +1,114 @@
...
+bool ControlBoardData::setInteractionMode(int axis, yarp::dev::InteractionModeEnum mode)
+{
+    yarp::sig::VectorOf<size_t> coupledActuatedAxes, coupledPhysicalJoints;
+    try
+    {
+        // If there is coupling let's check if we have to change the interaction mode for all the coupled joints
+        if (this->ijointcoupling)
+        {
+            this->ijointcoupling->getCoupledActuatedAxes(coupledActuatedAxes);
+            this->ijointcoupling->getCoupledPhysicalJoints(coupledPhysicalJoints);
```
traversaro, line 15: Can we cache this in an attribute buffer instead of calling it every time?

plugins/controlboard/src/ControlBoard.cpp
```diff
@@ -241,10 +260,27 @@ bool ControlBoard::setJointProperties(EntityComponentManager& _ecm)
             gzJoint.EnableTransmittedWrenchCheck(_ecm, true);
 
             // Initialize JointProperties object
-            m_controlBoardData.joints[i].name = jointFromConfigName;
+            m_controlBoardData.physicalJoints[i].commonJointProperties.name = jointFromConfigName;
 
             yInfo() << "Joint " << jointFromConfigName << " added to the control board data.";
         }
+        // Let's initialize all the buffers/vectors
+        configureBuffers();
```
traversaro, line 268: This function can fail, can we propagate the error to the caller?

traversaro (changes requested):
Minor comments, the one that I care more is to avoid to change the content of `m_pluginParameters`

plugins/controlboard/src/ControlBoardData.cpp
```diff
@@ -0,0 +1,114 @@
+
+#include "ControlBoardData.hh"
```
xela-95, line 2: Just to be consistent with the codebase

```suggestion
#include <ControlBoardData.hh>
```

plugins/controlboard/include/ControlBoardData.hh
```diff
@@ -66,11 +76,15 @@ class ControlBoardData
 {
 public:
     std::mutex mutex;
-    std::vector<JointProperties> joints;
+    std::vector<PhysicalJointProperties> physicalJoints;
+    std::vector<ActuatedAxesProperties>  actuatedAxes;
     yarp::os::Stamp simTime;
-
+    yarp::dev::IJointCoupling* ijointcoupling{nullptr};
```
xela-95, line 82: Can we try to use a smart pointer like `std::shared_ptr` instead of raw pointers?

plugins/controlboard/src/ControlBoard.cpp
```diff
@@ -364,13 +422,15 @@ bool ControlBoard::updateTrajectories(const UpdateInfo& _info, EntityComponentMa
...
-
-    for (auto& joint : m_controlBoardData.joints)
+    
+    for (auto& joint : m_controlBoardData.actuatedAxes)
     {
-        switch (joint.controlMode)
+        
+        switch (joint.commonJointProperties.controlMode)
         {
+        // PUT THE COUPLING HERE!!
```
xela-95, line 440: Leftover?
```suggestion
```

plugins/controlboard/include/ControlBoardData.hh
Nicogene: mmm I am not a great fan of using smart pointers w/ YARP interfaces, because the memory ownership when you do the `view` is not very clear to me

xela-95 (approved):
Thanks a lot @Nicogene for the great work! It looks good to me! 🚀 

Maybe it could be useful to add the coupled pendulum tutorial as suggested by @traversaro both as an example for the user and as a smoke test for the coupling feature. But this can be done in another PR.

plugins/controlboard/include/ControlBoardData.hh
traversaro: Yes, I can confirm. If you need to access a pointer that is not managed as smart point, you can't use smart pointers.

plugins/controlboard/src/ControlBoard.cpp
traversaro: Ack, can you add a comment explaining that and avoid to change the content of `m_pluginParameters` and create a new `yarp::os::Property` instance instead? Thanks!

plugins/controlboard/include/ControlBoardData.hh
xela-95: Ah ok thanks understood now!

Nicogene:
I should have addressed all your requested changes, and I re-tested the coupling after the changes

traversaro (approved):
Great, thanks a lot!

traversaro:
@xela-95 the PR for me is ready to merge once CI is happy. I guess we can merge with squash as the history is not curated.

PR #228, opened 2025-01-09
Add coupled_pendulum tutorial and tests

martinaxgloria:
This PR adds the `coupled_pendulum` tutorial. Tested:

https://github.com/user-attachments/assets/dea7a9a4-2e00-45a8-a647-1da1f66fba06

It's a draft PR since the corresponding unit test is still missing.

cc @xela-95 @traversaro @Nicogene

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ project(gz-sim-yarp-plugins
 
 find_package(YARP REQUIRED COMPONENTS robotinterface os)
 find_package(YCM REQUIRED)
+find_package(ICUB QUIET)
 
 # Initial value of GZ_SIM_YARP_PLUGINS_USED_GZ_SIM_VERSION is empty, and then is set to 8 or 9 depending if gz-sim8 or gz-sim9 are available
 set(GZ_SIM_YARP_PLUGINS_USED_GZ_SIM_VERSION_DOCS "Version of gz-sim used to compile gz-sim-yarp-plugins (either 8 or 9)")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,6 @@ project(gz-sim-yarp-plugins
 
 find_package(YARP REQUIRED COMPONENTS robotinterface os)
 find_package(YCM REQUIRED)
-find_package(ICUB QUIET)
 
 # Initial value of GZ_SIM_YARP_PLUGINS_USED_GZ_SIM_VERSION is empty, and then is set to 8 or 9 depending if gz-sim8 or gz-sim9 are available
 set(GZ_SIM_YARP_PLUGINS_USED_GZ_SIM_VERSION_DOCS "Version of gz-sim used to compile gz-sim-yarp-plugins (either 8 or 9)")
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,4 +1,13 @@
 find_package(GTest REQUIRED)
+# Only if the GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN variable is not defined,
+# we look for ICUB, to set its default value
+if(NOT DEFINED GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN)
+  find_package(ICUB QUIET)
+endif()
+option(GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN "Enable tests that require the installation of the icub-main package" ICUB_FOUND)
+if (GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN)
+  find_package(ICUB REQUIRED)
+endif()
 
 add_subdirectory(forcetorque)
 add_subdirectory(imu)
--- a/tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt
+++ b/tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt
@@ -17,6 +17,11 @@ gsyp_add_tutorial_model_check(forcetorque-model-one-sensor forcetorque/model_one
 gsyp_add_tutorial_model_check(forcetorque-model-two-sensors forcetorque/model_two_sensors/model2sensors.sdf)
 gsyp_add_tutorial_model_check(single-pendulum single_pendulum/model.sdf)
 
+
+if(GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN)
+    gsyp_add_tutorial_model_check(coupled-pendulum coupled_pendulum/model.sdf)
+endif()
+
 # Then check that the command correctly exit a failure if we pass it a invalid model
 add_test(NAME gz-sim-yarp-plugins-check-model-fails-as-expected-with-wrong-world-file
          COMMAND gz-sim-yarp-plugins-check-model --world-file  ./world_with_robotinterface_with_non_existing_filename.sdf
--- a/tutorial/coupled_pendulum/README.md
+++ b/tutorial/coupled_pendulum/README.md
@@ -1,0 +1,38 @@
+# Run coupled pendulum tutorial in Gazebo
+
+## Run model in Gazebo with YARP integration
+
+- 1st terminal:
+
+  ~~~bash
+  yarp server
+  ~~~
+
+- 2nd terminal:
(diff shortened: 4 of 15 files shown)


CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ project(gz-sim-yarp-plugins
 
 find_package(YARP REQUIRED COMPONENTS robotinterface os)
 find_package(YCM REQUIRED)
+find_package(ICUB QUIET)
```
traversaro, line 9: Why we added this?

tutorial/coupled_pendulum/README.md
```diff
@@ -0,0 +1,38 @@
+# Run coupled pendulum tutorial in Gazebo
+
```
traversaro, line 2: Somewhere here we should document that this model has a runtime dependency on `icub-main`.

CMakeLists.txt
Nicogene: My bad, I said to @martinaxgloria to add find_package but actually since it is a run-time and not compile-time dependency this is not needed

traversaro:
> It's a draft PR since the corresponding unit test is still missing.

If you do not have time to add a unit test, we can also just add a tutorial. Let's avoid that further enhancement prevent us to get improvements (see https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good).

martinaxgloria:
> If you do not have time to add a unit test, we can also just add a tutorial. Let's avoid that further enhancement prevent us to get improvements (see [en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good)).

Hi @traversaro, sorry for the delay. I'm going to add the unit test by the end of the day, at the latest by tomorrow if you're fine

traversaro:
No problem, take your time, I just was going through the open issues and PRs.

martinaxgloria:
Hi @traversaro, at the end I added the coupled pendulum within the tests of the controlBoard (both position and position direct) and I also added the test for checking that the tutorial runs without errors. Since the tutorial depends on icub-main, I protected it with an option that is ON if icub-main is found, OFF otherwise. Please, let me know if it is the intended behavior, thanks!

cc @Nicogene @xela-95

traversaro:
There is something I do not understand. The `CheckPositionTrackingWithTrajectoryGenerationUsingPendulumModel` depends as well from ICUB, but it is not disabled by the CMake option. Nevertheless the CI tests (where ICUB is not installed) do not pass. Can you double check what is going on?

Can you also modify the PR title to mention "Add coupled_pendulum tutorial and tests"?

tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt
```diff
@@ -17,6 +17,16 @@ gsyp_add_tutorial_model_check(forcetorque-model-one-sensor forcetorque/model_one
 gsyp_add_tutorial_model_check(forcetorque-model-two-sensors forcetorque/model_two_sensors/model2sensors.sdf)
 gsyp_add_tutorial_model_check(single-pendulum single_pendulum/model.sdf)
 
+if(ICUB_FOUND)
+    option(ENABLE_COUPLED_PENDULUM_TEST "Enable coupled pendulum tests that require ICUB" ON)
+else()
+    set(ENABLE_COUPLED_PENDULUM_TEST OFF)
+endif()
```
traversaro, line 24: I think this logic can be useful for all tests, so I think we can move it upper in the tests CMakeList.txt (see the comment there).

traversaro, line 24:
```suggestion
```

tests/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 find_package(GTest REQUIRED)
+find_package(ICUB QUIET)
```
traversaro, line 2: Calling `find_package(ICUB QUIET)` all the time is a bit problematic as:
* A failing `find_package(ICUB QUIET)` call can be quite slow, as it checks for ICUB cmake config files in a lot of directories, so if the relative option is already define, we should not search for it.
* See https://github.com/ami-iit/bipedal-locomotion-framework/pull/930 for more details 

A possible improvement is:

```suggestion
# Only if the GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN variable is not defined,
# we look for ICUB, to set its default value
if(NOT DEFINED GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN)
  find_package(ICUB QUIET)
endif()
option(GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN "Enable tests that require the installation of the icub-main package" ICUB_FOUND)
if (GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN)
  find_package(ICUB REQUIRED)
endif()
```

tests/gz-sim-yarp-plugins-check-model/CMakeLists.txt
```diff
@@ -17,6 +17,16 @@ gsyp_add_tutorial_model_check(forcetorque-model-one-sensor forcetorque/model_one
...
+
+if(ENABLE_COUPLED_PENDULUM_TEST)
```
traversaro, line 26:
```suggestion
if(GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN)
```

See the other suggestions for the rationale. Furthermore, I would also add `-DGZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN:BOOL=ON` in the tests, so we avoid that for some reason ICUB is not available and the tests silently ignored instead of failing.

traversaro:
For what regards testing:
* The apt test requires to compile icub-main from source, so I would not enable icub-main tests there
* The conda-forge-ci test using micromamba is mostly superseded by the pixi test, so I would not update it.
* I think instead it make sense to add the icub-main dependency and the `-DGZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN:BOOL=ON` option to the pixi project (this will be automatically then used in CI)

If you need any in person clarification feel free to ask me.

xela-95 (approved):
Great! Thanks @martinaxgloria ! I have nothing to add to the @traversaro suggestions

martinaxgloria:
The CI is still failing due to:
```
Error:  |yarp.os.YarpPluginSettings| Cannot find "couplingICubEye" plugin (not built in, and no .ini file found for it)Check that YARP_DATA_DIRS leads to at least one directory with plugins/couplingICubEye.ini or share/yarp/plugins/couplingICubEye.ini in it
```

This is maybe because the [`couplingICubEye` PR](https://github.com/robotology/icub-main/pull/997) was done versus devel, so it doesn't appear within the master branch of icub-main. With @xela-95 we tried to add icub-main dependency in pixi.toml as:

```
icub-main = { git = "https://github.com/robotology/icub-main.git", rev = "devel"}
```

but pixi install failed locally with this error: 
```
Error:
  × failed to parse project from /home/mgloria/iit/robotology-superbuild/src/gz-sim-yarp-plugins/pixi.toml: source
  │ dependencies are used in the feature 'default', but the `pixi-build` preview feature is not enabled
```

cc @traversaro

traversaro:
Good point! We still need to release icub-main with coupling support. `pixi build` support for icub-main would be cool to have at some point, but it will probably take a bit more time. At this point if we are not going to release icub-main in a short time, we can disable the test for now and enable it again once icub-main with coupling support is released?

martinaxgloria:
> At this point if we are not going to release icub-main in a short time, we can disable the test for now and enable it again once icub-main with coupling support is released?

I don't know when the new tag will be released (cc @Nicogene @pattacini), but if you all agree for me it's ok to disable it until that moment

pattacini:
> I don't know when the new tag will be released (cc @Nicogene @pattacini), but if you all agree for me it's ok to disable it until that moment

Discussed F2F w/ @traversaro; let's wait for the Distro 2025.02 to tag icub-main.

martinaxgloria:
Ok, since we are waiting for the distro to tag icub-main we have two possibilities:

1) keep this PR open (I could put it in draft) until the new tag with the couplingICubEye will be released 
2) disabled the test so that we can close this PR and then open another one to re-enable it

traversaro:
Thanks! Good to go for me once the PR is happy. We will address eventual test problems once we enable the tests.

xela-95:
Thanks! Merging then 🚀

PR #230, opened 2025-02-05
Change depends_on with depends-on in pixi.toml

traversaro:
The `depends-on` syntax was supported since pixi 0.21 (see https://github.com/prefix-dev/pixi/pull/1310), so there risk of creating troubles to any user with an old pixi is non existent, as anyhow the lock file version that we used is newer if I am not wrong.


xela-95 (approved):
Thanks!

PR #234, opened 2025-02-10
Update .gitattributes to track `pixi.lock` in git-LFS

xela-95:
This PR makes the pixi.lock file being tracked by git-LFS. 

Note: since I do not want to update the history of the repository with commands like [`git lfs migrate`](https://github.com/git-lfs/git-lfs/wiki/Tutorial#migrating-existing-repository-data-to-lfs) the size of the repo will not decrease with this PR, but at least from now on all the updates will be stored in LFS.


traversaro:
I would strongly prefer not do to that. Differently from `ami-iit`, I am not aware of any Git LFS voucher being active on the `robotology` organization, and even if  I would like to avoid having to worry about that voucher being valid/active. Just to understand, did you ever encountered any problem due to the size of the repo?

xela-95:
> I would strongly prefer not do to that. Differently from `ami-iit`, I am not aware of any Git LFS voucher being active on the `robotology` organization, and even if I would like to avoid having to worry about that voucher being valid/active. Just to understand, did you ever encountered any problem due to the size of the repo?

Ok for me! No I just thought we forgot about this and I was worried that in the future, by adding revisions of the pixi.lock the dimension of the repo would increase without big benefits. Closing then!

traversaro:
Thanks!

Issue #235, opened 2025-02-12
Implement ISystemReset interface in all plugins and ensure that a system reset does not result in YARP ports being closed and reopened


traversaro:
See https://github.com/robotology/gz-sim-yarp-plugins/issues/206#issuecomment-2652966078 .

LoreMoretti:
It would be extremely useful that when the simulation gets reset the robot will start from the initial configuration set through the `gzyarp::ConfigurationOverride` plugin in the world file.

traversaro:
> It would be extremely useful that when the simulation gets reset the robot will start from the initial configuration set through the `gzyarp::ConfigurationOverride` plugin in the world file.

It does not? That's a bug that I think is independent from the one tracked here.

traversaro:
> > It would be extremely useful that when the simulation gets reset the robot will start from the initial configuration set through the `gzyarp::ConfigurationOverride` plugin in the world file.
> 
> It does not? That's a bug that I think is independent from the one tracked here.

Follow up in https://github.com/robotology/gz-sim-yarp-plugins/issues/252 .

Issue #206, opened 2024-08-29
GUI Crash when resetting the simulation


LoreMoretti:
When I reset the simulation the GUI crashes as per the video below:

https://github.com/user-attachments/assets/e09f64cb-f557-4253-9afe-f5f1934671e4

It follows the log after the reset is invoked:

```

[INFO] gz-sim-yarp-clock-system plugin Reset invoked.
[WARNING] No initial configuration found, initializing trajectory generator with current values
[WARNING] No initial configuration found, initializing trajectory generator with current values
[WARNING] No initial configuration found, initializing trajectory generator with current values
[WARNING] No initial configuration found, initializing trajectory generator with current values
[INFO] Initial configuration found, initializing trajectory generator with it
[INFO] Initial configuration found, initializing trajectory generator with it
[INFO] interrupt1 phase starting...
[INFO] interrupt1 phase finished.
[INFO] shutdown phase starting...
[INFO] Entering action level 5 of phase shutdown
[INFO] Executing detach action, level 5 on device lidarWrapY with parameters []
[INFO] All actions for action level 5 of shutdown phase started. Waiting for unfinished actions.
[INFO] All actions for action level 5 of shutdown phase finished.
[INFO] Entering action level 15 of phase shutdown
[INFO] Executing detach action, level 15 on device left_arm-mc_nws_yarp with parameters []
[INFO] Executing detach action, level 15 on device right_arm-mc_nws_yarp with parameters []
[INFO] Executing detach action, level 15 on device left_leg-mc_nws_yarp with parameters []
[INFO] Executing detach action, level 15 on device right_leg-mc_nws_yarp with parameters []
[INFO] Executing detach action, level 15 on device head-mc_nws_yarp with parameters []
[INFO] Executing detach action, level 15 on device torso-mc_nws_yarp with parameters []
[INFO] Executing detach action, level 15 on device left_leg-FT_wrapper with parameters []
[DEBUG] |yarp.device.multipleanalogsensorsserver| Detach complete
[INFO] Executing detach action, level 15 on device right_leg-FT_wrapper with parameters []
[DEBUG] |yarp.device.multipleanalogsensorsserver| Detach complete
[INFO] Executing detach action, level 15 on device left_arm-FT_wrapper with parameters []
[DEBUG] |yarp.device.multipleanalogsensorsserver| Detach complete
[INFO] Executing detach action, level 15 on device right_arm-FT_wrapper with parameters []
[...]

PR #236, opened 2025-02-12
Remove all reset methods in Clock, ConfigurationOverride and ControlBoard plugins

traversaro:
Defining the `ISystemReset` interface in some plugins and not in others create a problematic situation in which some plugins are restarted in a way (calling the destructor and constructor) and some other in another way (calling the reset method). This resulted in unexpected behaviour as the one reported in https://github.com/robotology/gz-sim-yarp-plugins/issues/206 .

To solve at least the crash, this PR removes all the reset methods. The downside of this is that now all the YARP ports are closed and re-opened at each reset, but anyhow this is better then just crash. 

Implementing a smarter reset strategy that does not restart the ports (that by the way is how Gazebo Classic and gazebo-yarp-plugins worked) will be tracked in https://github.com/robotology/gz-sim-yarp-plugins/issues/235 .

### DIFF
--- a/plugins/clock/Clock.cc
+++ b/plugins/clock/Clock.cc
@@ -32,7 +32,7 @@ using yarp::os::BufferedPort;
 namespace gzyarp
 {
 
-class Clock : public System, public ISystemConfigure, public ISystemPostUpdate, public ISystemReset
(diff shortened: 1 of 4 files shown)


xela-95 (approved):
Thanks! LGTM

PR #237, opened 2025-02-12
Remove warning and add pragmatic compatibility policy

traversaro:
Now that Gazebo Classic has reached EOL and more users started to use `gz-sim-yarp-plugins`, I think we reached the time to define a pragmatic compatibility policy. I tried to draft a simple yet useful policy, in particular:
* We try to reserve breaking changes (remove of parameters) in major releases, but we reserve the possibility to do breaking changes even in minor releases if it is necessary.
* We promise not to touch public headers in patch releases. This is useful to ensure that we do not break ABI in patch releases, without the need to actually caring about what is an ABI change and what not. I would like to have this to solve https://github.com/conda-forge/gz-sim-yarp-plugins-feedstock/issues/17 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,12 @@
 # gz-sim-yarp-plugins
 
-> [!WARNING]
-> This repository contains a preliminary work in progress of integration of Modern Gazebo (gz) and YARP devices, a port of some functionalities of <https://github.com/robotology/gazebo-yarp-plugins> to Modern Gazebo. \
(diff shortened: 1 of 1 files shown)


traversaro:
For any question or doubt please comment, thanks!

Nicogene (approved):

martinaxgloria (approved):

xela-95 (approved):

PR #240, opened 2025-02-25
Fix compilation with YARP master branch (future YARP 3.12) as of 2025/02/25

traversaro:
This PR switches migrate the Laser plugin to the use of yarp::dev::ReturnValue-based interface, as a follow up of:
* https://github.com/robotology/yarp/discussions/3168
* https://github.com/robotology/yarp/pull/3180
* https://github.com/robotology/yarp-device-rplidar/pull/7

To avoid the need to keep two branches, one for compatibility with YARP 3.11.* and one for the future YARP 3.12.*, I prepared a `YarpDevReturnValueCompat.h` header that can be useful to implement an interface that works in both cases. As these header can be useful also for other repos, but I do not think it make sense to install in an a reusable library as it is a temporary solution, I will store it in https://gist.github.com/traversaro/ab7ab377c70c50d312a7b5edd4da6242, so that interested users can copy it in their repo as a private (non-installed) header to simplify the support of the future YARP 3.12

### DIFF
--- a/libraries/common/CMakeLists.txt
+++ b/libraries/common/CMakeLists.txt
@@ -1,3 +1,5 @@
+# This library is installed as it is a shared library, but no headers are installed,
+# so the headers are effectively private headers, even if the library they belong to is installed
 add_library(gz-sim-yarp-commons SHARED Common.hh ConfigurationHelpers.hh ConfigurationHelpers.cpp)
 
 target_include_directories(gz-sim-yarp-commons PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>")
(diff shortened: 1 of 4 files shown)


xela-95 (approved):
Thanks @traversaro !

xela-95:
There's just the CI action on Ubuntu 22.04 failing due to some reason that seems unrelated to this specific PR.

traversaro:
> There's just the CI action on Ubuntu 22.04 failing due to some reason that seems unrelated to this specific PR.

I think we can track that failure (that also happens without this PR) in https://github.com/robotology/gz-sim-yarp-plugins/issues/241 , so I think we can safely merge if you agree.

xela-95:
> failure

perfect thanks!

Issue #245, opened 2025-03-04
Torque control mode: cannot set reference torque


PasMarra:
https://github.com/user-attachments/assets/d8369dff-a3c2-4ada-94bc-260d70b67416

When switching to torque control mode, the joint does not keep the measured torque as reference. Furthermore, it is not possible to set a new reference.

traversaro:
> When switching to torque control mode, the joint does not keep the measured torque as reference.

Thanks! That is probably a bug in the switching of control modes. The related code in gazebo-yarp-plugins is https://github.com/robotology/gazebo-yarp-plugins/blob/7206f84eca17d83f6b9d5b90e3f1b05d0be7a78e/plugins/controlboard/src/ControlBoardDriverControlMode.cpp#L139-L181 , but probably there is no equivalent code in gz-sim-yarp-plugins. fyi @xela-95 

> Furthermore, it is not possible to set a new reference.

In which control mode? Can you point to the part of the video where this happens?

PasMarra:
> In which control mode? 

Torque control

> Can you point to the part of the video where this happens?

`shoulder_roll`:
1- Switching to torque control mode: min 0:06.
2- Setting new reference for the torque control mode: min 0:09 - 0:15.

`shoulder_pitch`:
1- Switching to torque control mode: min 0:21.
2- Setting new reference for the torque control mode: min 0:24 - 0:28.

traversaro:
Thanks a lot @PasMarra for reporting this critical regression, this may be of interest for other users of gz-sim @LoreMoretti . As the main problem was the torque reference not being correctly set, I would split the other problem ("When switching to torque control mode, the joint does not keep the measured torque as reference. ") in a separate issue: https://github.com/robotology/gz-sim-yarp-plugins/issues/249 .

Issue #247, opened 2025-03-04
VOCAB_CM_TORQUE control not supported in coupled joints


traversaro:
The code that converts desired quantities from actuated axis to physical joints and measured quantities from physical joints to actuated axis seems only consider desired positions and measured position and velocities, see:
* https://github.com/robotology/gz-sim-yarp-plugins/blob/13720141e6f95574d83042db1a5572298c44085f/plugins/controlboard/src/ControlBoard.cpp#L382-L390
* https://github.com/robotology/gz-sim-yarp-plugins/blob/13720141e6f95574d83042db1a5572298c44085f/plugins/controlboard/src/ControlBoard.cpp#L472-L481

So, clearly the `VOCAB_CM_TORQUE` control mode is not supported for coupled joints. At the moment we should document this limitation, and ideally return an error if someone asks to set in torque a coupling joint. We could also try to understand if it make sense to support this (for example if it is supported in any physical robot) and if it exists a case in which it make sense to support this, implement it.

traversaro:
> I guess the simplest solution was to implement it: [#248](https://github.com/robotology/gz-sim-yarp-plugins/pull/248) .

This works fine for coupling devices that implement the torque mapping methods, but some do not (see https://github.com/robotology/gz-sim-yarp-plugins/pull/248#pullrequestreview-2660646495). In that case we need to understand how to handle. For the control I think the best solution is return error if somebody tries to switch the coupled joints in torque control mode, while I am not sure how to handle the torque feedback case. In theory we could return false in `getTorques`, but I am afraid this would not be handled nicely by the various remappers logic. Probably the easiest solution is to return the same value returned in the real robot for the ergocub heands, @PasMarra as you are probably the person with most experience with the real robot, can you check that? Thanks!

traversaro:
Another problem in the torque control mode in coupled joints is that given how it is implemented now is possible that in underactuated systems that approximate system in which there is come kind of "position constraint" among the different physical joints, controlling the coupled device in torque will result in a drift that will break the position constraints. This does not happen in other control modes as there eventually everything is converted to a reference position on the actuated axis side, that is converted via position mapping to a physical joint reference that then is passed to a position PID, so the enforcing of any position constraint is done implicitly by the PID position control on the physical joint side.

PasMarra:
Hi @traversaro, I've tried  the torque control mode for the fingers of ergoCub002, but also on the real robot this control mode is not properly handled.

[...]

Issue #241, opened 2025-02-26
Periodic conda CI failed February 2025 with gcc error error: 'timespec_get' has not been declared in '::'    80 |   using ::timespec_get;


traversaro:
Last working build: https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/13468652828 .
First failing build: https://github.com/robotology/gz-sim-yarp-plugins/actions/runs/13478840641 .

Error:

~~~
FAILED: plugins/laser/CMakeFiles/gz-sim-yarp-laser-system.dir/Laser.cc.o 
[...]

PR #248, opened 2025-03-05
Fix setRefTorque doing nothing since 0.4.0 and torque feedback in coupled case when torque mapping methods are implemented

traversaro:
https://github.com/robotology/gz-sim-yarp-plugins/pull/221 contained a critical regression that got released in 0.4.0. Due to the coupling refactor, even in torque control mode calling `setRefTorque` was ignored, as the value `m_controlBoardData.actuatedAxes[i].commonJointProperties.refTorque` (that was set by `setRefTorque` was never copied in `m_controlBoardData.physicalJoints[i].commonJointProperties.refTorque` (that was the value eventually passed to the Gazebo simulation. The problem was detected and reported by @PasMarra in https://github.com/robotology/gz-sim-yarp-plugins/issues/245 .

This PR fixes this bug, and also:
* Fixes the `ControlBoardTorqueControlTest` to also check if the value set by `setRefTorque` is correctly propagated. The test was already using `setRefTorque`, but then it was not checking if the value set by `setRefTorque` was the value read by `getTorque`, so I added the check ` EXPECT_NEAR(jointTorque, motorTorque, acceptedTolerance);` to prevent further regression in the future. I also cleaned up some old debug prints, and reduced the iterations of the test from 1000 to 10, to reduce the time of the test from ~3 seconds to ~2 seconds (that it is still a lot, but that is another story). 
* Ensure that the torque reference and the torque feedback is correctly propagated also for coupled joints in the case the coupling exposes the required torque-related mapping, fixing part of https://github.com/robotology/gz-sim-yarp-plugins/issues/247 . See https://github.com/robotology/gz-sim-yarp-plugins/pull/248#pullrequestreview-2660646495 and https://github.com/robotology/gz-sim-yarp-plugins/pull/248#issuecomment-2700518363 for more details.

Fix https://github.com/robotology/gz-sim-yarp-plugins/issues/245

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.22)
 
 project(gz-sim-yarp-plugins
         LANGUAGES CXX C
-        VERSION 0.4.0)
+        VERSION 0.4.1)
 
 find_package(YARP REQUIRED COMPONENTS robotinterface os)
 find_package(YCM REQUIRED)
--- a/plugins/controlboard/src/ControlBoard.cpp
+++ b/plugins/controlboard/src/ControlBoard.cpp
@@ -479,7 +479,7 @@ bool ControlBoard::updateReferences(const UpdateInfo& _info, EntityComponentMana
         }
 
         m_controlBoardData.ijointcoupling->convertFromActuatedAxesToPhysicalJointsPos(m_actuatedAxesPositionBuffer, m_physicalJointsPositionBuffer);
-        m_controlBoardData.ijointcoupling->convertFromActuatedAxesToPhysicalJointsTrq(m_actuatedAxesTorqueBuffer, m_physicalJointsTorqueBuffer);
+        m_controlBoardData.ijointcoupling->convertFromActuatedAxesToPhysicalJointsTrq(m_actuatedAxesPositionBuffer, m_actuatedAxesTorqueBuffer, m_physicalJointsTorqueBuffer);
 
         for(auto i = 0; i < m_controlBoardData.physicalJoints.size(); i++)
         {
--- a/plugins/controlboard/src/ControlBoard.cpp
+++ b/plugins/controlboard/src/ControlBoard.cpp
@@ -366,7 +366,6 @@ bool ControlBoard::readJointsMeasurements(const gz::sim::EntityComponentManager&
 
         if (gzJoint.TransmittedWrench(_ecm).has_value())
         {
-            // joint.commonJointProperties.torque = gzJoint.TransmittedWrench(_ecm).value().at(0).torque().x();
             joint.commonJointProperties.torque
                 = getJointTorqueFromTransmittedWrench(gzJoint,
                                                       gzJoint.TransmittedWrench(_ecm).value().at(0),
@@ -382,10 +381,13 @@ bool ControlBoard::readJointsMeasurements(const gz::sim::EntityComponentManager&
     if (m_controlBoardData.ijointcoupling){
         m_controlBoardData.ijointcoupling->convertFromPhysicalJointsToActuatedAxesPos(m_physicalJointsPositionBuffer, m_actuatedAxesPositionBuffer);
         m_controlBoardData.ijointcoupling->convertFromPhysicalJointsToActuatedAxesVel(m_physicalJointsPositionBuffer, m_physicalJointsVelocityBuffer, m_actuatedAxesVelocityBuffer);
+        m_controlBoardData.ijointcoupling->convertFromPhysicalJointsToActuatedAxesTrq(m_physicalJointsPositionBuffer, m_physicalJointsTorqueBuffer, m_actuatedAxesTorqueBuffer);
+
         for (int i = 0; i < m_controlBoardData.actuatedAxes.size(); i++)
         {
             m_controlBoardData.actuatedAxes[i].commonJointProperties.position = m_actuatedAxesPositionBuffer[i];
             m_controlBoardData.actuatedAxes[i].commonJointProperties.velocity = m_actuatedAxesVelocityBuffer[i];
+            m_controlBoardData.actuatedAxes[i].commonJointProperties.torque = m_actuatedAxesTorqueBuffer[i];
         }
     }
     else {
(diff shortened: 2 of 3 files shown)


traversaro:
@Nicogene @xela-95 the regression in setRefTorque is quite critical and problematic for users of gz-sim on the AMI side, so if you agree I would quickly release a 0.4.1 after this PR has been reviewed and merged.

traversaro:
The PR is ready for review. The only failure in CI is https://github.com/robotology/gz-sim-yarp-plugins/issues/241, that is already tracked there.

Nicogene:
> @Nicogene @xela-95 the regression in setRefTorque is quite critical and problematic for users of gz-sim on the AMI side, so if you agree I would quickly release a 0.4.1 after this PR has been reviewed and merged.

Agree!

Nicogene (commented):
Since we didn't find a mathematical representation from converting the torque from the actuated axes to physical joints domain (and vice-versa, see [here](https://mathb.in/76391)), in `couplingXCubHandMk5` (and in other coupling devices), it does nothing, it just return `false`, so I think that the torque control would not work when the coupling is involved

traversaro:
Actually I wonder if we can keep the coupling change also for torque, as anyhow it is a net improvement. Both before and after this PR, for devices that not implement `convertFromPhysicalJointsToActuatedAxesTrq` and `convertFromActuatedAxesToPhysicalJointsTrq` the torque control silently fails (so it is exactly the same behaviour), but this PR at least fixes the case for devices that do implement `convertFromPhysicalJointsToActuatedAxesTrq` and `convertFromActuatedAxesToPhysicalJointsTrq`,so it seems to me a net positive, even if it does not solve completely https://github.com/robotology/gz-sim-yarp-plugins/issues/247 . What do you think @Nicogene ?

traversaro:
> (and vice-versa, see [here](https://mathb.in/76391))

Unrelated, but that website is shutting down, let's backup the script in here.

<details>

<summary>Click to see the linked pastebin</summary>

## Theory

$$
\theta \in \mathbb{R}^{m} : \text{Actuated Axes positions}
$$

$$
q \in \mathbb{R}^{n} : \text{Physical Joints positions}
$$

In general, we have $m \leq n$. 

There is unique value of Actuated Axes corresponding to a given Physical Joint position, while in general there may be multiple Physical Joint position corresponding to a given Actuated Axes. 

So, the mapping from Physical Joint $q$ to Actuated Axes $\theta$ is defined as:

$$
\theta = f(q)
$$

while the mapping from Actuated Axes $\theta$  to Physical Joint $q$  is defined as:

$$
q = g(\theta)
$$

We have the following property:

$$
f(g(\theta)) = \theta
$$

while in general it is not always true that the inverse holds, i.e. 

$$
g(f(q)) \neq q
$$


## Method Definitions for methods to convert from Physical Joints to Actuated Axis values

In the rest of the section, we assume that $\theta(t) = f(q(t))$ and $\dot{\theta}(t)^T \tau_{\theta}(t) = \dot{q}(t)^T \tau_{q}(t)$

### `convertFromPhysicalJointPosToActuatedAxisPos`

This method implements $f(q)$

### `convertFromPhysicalJointPosToActuatedAxisVel`

This method implements $f'$ that can be used to compute $\dot{\theta}(t)$, defined such that:

$$
\dot{\theta}(t) = \frac{\partial}{\partial q} f(q(t)) \dot{q}(t) = f'( q(t), \dot{q}(t))
$$

### `convertFromPhysicalJointPosToActuatedAxisAcc`

This method implements $f''$ that can be used to compute $\ddot{\theta}(t)$,  defined such that:


$$
\ddot{\theta}(t) = \frac{\partial}{\partial q} f'(q(t), \dot{q}(t)) \dot{q}(t) + \frac{\partial}{\partial \dot{q}} f'(q(t), \dot{q}(t)) \ddot{q}(t) = f''( q(t), \dot{q}(t), \ddot{q}(t))
$$

### `convertFromPhysicalJointPosToActuatedAxisTrq`

This method implements $\tilde{f}$ that can be used to compute $\tau_{\theta}(t)$, defined as:

??? . I have no idea on how this can be defined at the moment.



## Method Definitions for methods to convert from Actuated Axis to Physical Joints values

### `convertFromActuatedAxisToPhysicalJointPos`

This method implements (a possible choice of) $g(\theta)$. Given the property defined earlier, there is no unique choice of $g(\theta)$ that respect the property $$f(g(\theta)) = \theta$$. A possible choise is to choose the one with minimum norm among the possible choices, but probably we may want to avoid constraining this as part of the interface definition.


### `convertFromActuatedAxisToPhysicalJointVel`


This method implements $g'$ that can be used to compute $\dot{q}(t)$, defined such that:

$$
\dot{q}(t) = \frac{\partial}{\partial \theta} g(\theta(t)) \dot{\theta}(t) = g'( \theta(t), \dot{\theta}(t))
$$


### `convertFromActuatedAxisToPhysicalJointAcc`


This method implements $f''$ that can be used to compute $\ddot{\theta}(t)$,  defined such that:


$$
\ddot{q}(t) = \frac{\partial}{\partial \theta} g'(\theta(t), \dot{\theta}(t)) \dot{\theta}(t) + \frac{\partial}{\partial \dot{\theta}} g'(\theta(t), \dot{\theta}(t)) \ddot{\theta}(t) = g''( \theta(t), \dot{\theta}(t), \ddot{\theta}(t))
$$

### `convertFromActuatedAxisToPhysicalJointTrq`

??? . I have no idea on how this can be defined at the moment.

</details>

traversaro:
By the way, I wonder why I was so confused, the co-adjoin mapping is well defined in that case, it is the mapping that preserve the inner product between velocity and torque (i.e. the instantenous power), if anyone is interested I can show him with a whiteboard.

Nicogene:
> Actually I wonder if we can keep the coupling change also for torque, as anyhow it is a net improvement. Both before and after this PR, for devices that not implement `convertFromPhysicalJointsToActuatedAxesTrq` and `convertFromActuatedAxesToPhysicalJointsTrq` the torque control silently fails (so it is exactly the same behaviour), but this PR at least fixes the case for devices that do implement `convertFromPhysicalJointsToActuatedAxesTrq` and `convertFromActuatedAxesToPhysicalJointsTrq`,so it seems to me a net positive, even if it does not solve completely #247 . What do you think @Nicogene ?

Ok for me!

Nicogene (approved):

PR #250, opened 2025-03-05
Cleanup libGL conda dependencies

traversaro:
Since the `libgl` library was packaged in conda-forge, we can now just depend on it (and the related system-level OpenGL drivers, installed via apt) instead of rely on CDTs packages, see https://conda-forge.org/docs/maintainer/maintainer_faq/#how-do-i-fix-the-libglso1-import-error for upstream conda-forge documentation.

### DIFF
--- a/.github/workflows/conda-forge.yml
+++ b/.github/workflows/conda-forge.yml
@@ -36,12 +36,12 @@ jobs:
       shell: bash -l {0}
       run: |
         # Additional dependencies only useful on Linux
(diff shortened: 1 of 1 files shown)


xela-95 (approved):

traversaro:
Cool, this solved https://github.com/robotology/gz-sim-yarp-plugins/issues/241 ! Indeed I was looking into the CI to investigate the problem, but I would have not expected to find it at the first try.

PR #251, opened 2025-03-06
Changed scope for os and robotinterface in libraries/common

elandini84:
YARP_os and YARP_robotinterface are now linked with PUBLIC scope to allow compile the repo when yarp is not installed


xela-95 (approved):
Thanks! It looks good to me

traversaro (approved):
Thanks a lot @elandini84 !

Issue #252, opened 2025-03-10
Allow to restart from initial configuration when resetting the simulation


LoreMoretti:
I am using [this version](https://github.com/robotology/gz-sim-yarp-plugins/tree/9691fa7f632172d93890c104167d3f6828749cb8) of `gz-sim-yarp-plugins`.

What I usually do is:

1. run `yarpserver`
2. start gz-sim like `YARP_CLOCK=/clock gz sim -r ./world/world.sdf `
3. start `WBD`
4. run a controller.

Then, I do:

1. stop the controller
2. reset the gz simulation (`ctrl+R` or clicking the reset icon)

But, I noticed that the robot it is not placed back to the initial configuration specified through the `gzyarp::ConfigurationOverride` plugin in the world file.

cc @traversaro

traversaro:
Ok, but the initial configuration specified from the `gzyarp::ConfigurationOverride` works fine the first time, right?

LoreMoretti:
> Ok, but the initial configuration specified from the `gzyarp::ConfigurationOverride` works fine the first time, right?

yes, it works fine.

traversaro:
I was able to reproduce the problem in the case of https://github.com/robotology/gz-sim-yarp-plugins/blob/main/tutorial/single_pendulum/single_pendulum_world.sdf, but I had to add to the world also the clock plugin and to set `YARP_CLOCK=/clock`.

traversaro:
However, I am not sure if the problem of `YARP_CLOCK=/clock gz sim -r ./world/world.sdf` is that the initial position is not reset. It seems to me that if I launch  `YARP_CLOCK=/clock gz sim -r ./world/world.sdf` and then I try to reset, the simulation does not restart? Probably in https://github.com/robotology/gz-sim-yarp-plugins/pull/236 I was a bit too extremistic, and instead it make sense to have the reset method, at least for the Clock plugins so that the port of the clock is never closed? Anyhow, the CLock plugin is not related at all with the robotinterface, so I guess that it is not a problem if only that one has a reset properly implemented.

traversaro:
Even implementing the reset, during reset the gz sim process blocks after the message:

~~~
[INFO] Executing detach action, level 5 on device single_pendulum_nws_yarp with parameters []
~~~

and the following message never appears:

~~~
[INFO] All actions for action level 5 of shutdown phase started. Waiting for unfinished actions.
~~~

[...]

Issue #215, opened 2024-10-04
Test fails with segfault at exit on Windows [Ionic]


traversaro:
.

traversaro:
See https://github.com/robotology/gz-sim-yarp-plugins/issues/205#issuecomment-2393288115 .

PR #253, opened 2025-03-17
Make sure that gzyarp::RobotInterface plugin is always closed before gzyarp::Clock destruction and do not run macOS tests in CI

traversaro:
Furthermore, to avoid deadlocks due to the use of YARP_CLOCK environment variable, while the gzyarp::Clock destruction is ongoing, spawn a thread to increase the value published on the /clock port to avoid that any yarp::os::PeriodicThread spawned by the gzyarp::RobotInterface is blocked in a deadlock.

This solution is not ideal as it publishes some spurious non-simulated clock instances on the `/clock` port and it spawns a thread in the Clock destructors, but unfortunately is is the only solution possible without changing the clock protocol that is assumed by the yarp::os::NetworkClock class.

To avoid regressions, these PR also add a test of a model that uses both `gzyarp::RobotInterface` and `gzyarp::Clock` plugins, and as part of the same test the behaviour in case of `Reset` is also tested.

Fix https://github.com/robotology/gz-sim-yarp-plugins/issues/252


traversaro:
Interesting, the tests are sometimes timeouting, let me try locally.

plugins/robotinterface/RobotInterface.cc
```diff
@@ -65,13 +65,20 @@ class RobotInterface : public System, public ISystemConfigure
...
+            m_deviceRemovedConnection.reset();
+            m_clockPluginRemovedConnection.reset();
             m_robotInterfaceCorrectlyStarted = false;
         }
     }
 
-    void OnDeviceRemoved(std::string removeDeviceRegistryDatabaseKey)
+    void OnDeviceRemoved(std::string gzInstanceId, std::string removeDeviceRegistryDatabaseKey)
     {
+        // If th device removed is not in the same gz instance of the robotinterface, return
```
xela-95, line 76:
```suggestion
        // If the device removed is not in the same gz instance of the robotinterface, return
```

plugins/robotinterface/RobotInterface.cc
```diff
@@ -82,6 +89,25 @@ class RobotInterface : public System, public ISystemConfigure
         return;
     }
 
+    void OnClockPluginRemoved(std::string gzInstanceId, std::string removedClockPluginID)
+    {
+        // If th device removed is not in the same gz instance of the robotinterface, return
```
xela-95, line 94:
```suggestion
        // If the device removed is not in the same gz instance of the robotinterface, return
```

tests/clock/ClockTest.cc
```diff
@@ -98,11 +102,68 @@ TEST_F(ClockTestFixture, SimulationStartsIfYARPClockAlreadySet)
...
     fixture.Server()->Run(/*_blocking=*/true, iterations, /*_paused=*/false);
 
     // ASSERT
     EXPECT_TRUE(yarp::os::Network::checkNetwork(1.0)) << "Error: YARP network not detected";
+
     // Check if the /clock port has been correctly created
     EXPECT_TRUE(yarp::os::NetworkBase::exists("/clock")) << "Error: /clock port does not exist";
+
+    // Reset
+    // Unfortunatly the reset can only be requested via a transport request
```
xela-95, line 145:
```suggestion
    // Unfortunately the reset can only be requested via a transport request
```

xela-95 (approved):
Great @traversaro !!

traversaro:
Ok, I found the problem. The problem is that in Gazebo Harmonic the robotinterface `parentEntityScopedName` somehow does not include the world. For example, with this debug print:

~~~diff
diff --git a/plugins/robotinterface/RobotInterface.cc b/plugins/robotinterface/RobotInterface.cc
index c0375a9..63099a2 100644
--- a/plugins/robotinterface/RobotInterface.cc
+++ b/plugins/robotinterface/RobotInterface.cc
@@ -91,6 +91,8 @@ public:

     void OnClockPluginRemoved(std::string gzInstanceId, std::string removedClockPluginID)
     {
+
+        std::cerr << "============> OnClockPluginRemoved " << gzInstanceId << " " << removedClockPluginID  << "  " << m_parentEntityScopedName << std::endl;
         // If th device removed is not in the same gz instance of the robotinterface, return
         if (gzInstanceId != m_gzInstanceId)
         {
~~~

on Harmonic in ClockTest I get:

~~~
15: ============> OnClockPluginRemoved 0x55bedf6cd538 world/World/clockPluginPlaceholderInPlaceOfYarpDeviceName  model/single_pendulum
~~~

while in Ionic:

~~~
15: ============> OnClockPluginRemoved 0x55e5fcb2a4e8 world/World/clockPluginPlaceholderInPlaceOfYarpDeviceName  world/World/model/single_pendulum
~~~

In the Ionic case, the scoped name of the parent entity of the `gzyarp::RobotInterface` is the expected `world/World/model/single_pendulum`. This is the same problem documented in https://github.com/robotology/gz-sim-yarp-plugins/blob/ac44b8f186cc32b1a8995adaa32e1bc1df8c1981/plugins/configurationoverride/README.md?plain=1#L41, but we should report the issue upstream. In the meanwhile, we can just do not check for the parentScopedName in gz-sim8.

traversaro:
> In the meanwhile, we can just do not check for the parentScopedName in gz-sim8.

I did that in https://github.com/robotology/gz-sim-yarp-plugins/pull/253/commits/5d4c6b9eee078df634bd713185f30e27a492009b . To mitigate non-intuitive behaviours of robotinterface closing as a nested model with a clock plugin is deleted, I added the reference doc for the `gzyarp::Clock` plugin, explaining that the `gzyarp::Clock` plugin should only be used in top-level world files.

traversaro:
Ok, it seems that the fix was effective.

LoreMoretti:
I tested on my laptop. It works! Thanks @traversaro!

LoreMoretti (approved):

traversaro:
Cool, now the tests only fails on macOS. -_- 

I will test on the mac mini as soon as possible.

xela-95:
@traversaro I saw you also added documentation for the clock plugin, great! We should link it in the table of the main README

traversaro:
Interestingly, I thought I did that but there is no related commit, I will do it now.

Nicogene (approved):

traversaro:
Actually the Ionic tests on macOS were not running, I switched them to run on Ionic (and not on Harmonic) in https://github.com/robotology/gz-sim-yarp-plugins/issues/215, let's see if that works.

traversaro:
Indeed, the test is hanging also in Ionic.

traversaro:
The test also fails (sometime timeout, some other time segfault) on `Ionic`, and it seems some threading problem or similar related to gz-sensors:

˜˜˜
(lldb) bt
* thread #25, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000105d83ea8 libgz-sim9-sensors-system.9.0.0.dylib`std::__1::__hash_iterator<std::__1::__hash_node<std::__1::__hash_value_type<std::__1::reference_wrapper<std::type_info const>, std::__1::unique_ptr<gz::common::Event, std::__1::default_delete<gz::common::Event>>>, void*>*> std::__1::__hash_table<std::__1::__hash_value_type<std::__1::reference_wrapper<std::type_info const>, std::__1::unique_ptr<gz::common::Event, std::__1::default_delete<gz::common::Event>>>, std::__1::__unordered_map_hasher<std::__1::reference_wrapper<std::type_info const>, std::__1::__hash_value_type<std::__1::reference_wrapper<std::type_info const>, std::__1::unique_ptr<gz::common::Event, std::__1::default_delete<gz::common::Event>>>, gz::sim::v9::EventManager::Hasher, gz::sim::v9::EventManager::EqualTo, true>, std::__1::__unordered_map_equal<std::__1::reference_wrapper<std::type_info const>, std::__1::__hash_value_type<std::__1::reference_wrapper<std::type_info const>, std::__1::unique_ptr<gz::common::Event, std::__1::default_delete<gz::common::Event>>>, gz::sim::v9::EventManager::EqualTo, gz::sim::v9::EventManager::Hasher, true>, std::__1::allocator<std::__1::__hash_value_type<std::__1::reference_wrapper<std::type_info const>, std::__1::unique_ptr<gz::common::Event, std::__1::default_delete<gz::common::Event>>>>>::find<std::__1::reference_wrapper<std::type_info const>>(std::__1::reference_wrapper<std::type_info const> const&) + 160
    frame #1: 0x0000000105d6e2e4 libgz-sim9-sensors-system.9.0.0.dylib`void gz::sim::v9::EventManager::Emit<gz::common::EventT<void (), gz::sim::v9::events::RenderTeardownTag>>() + 60
    frame #2: 0x0000000105d6e10c libgz-sim9-sensors-system.9.0.0.dylib`gz::sim::v9::systems::SensorsPrivate::RenderThread() + 232
    frame #3: 0x0000000105d85dd0 libgz-sim9-sensors-system.9.0.0.dylib`void* std::__1::__thread_proxy[abi:ne180100]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (gz::sim::v9::systems::SensorsPrivate::*)(), gz::sim::v9::systems::SensorsPrivate*>>(void*) + 88
    frame #4: 0x0000000187e69f94 libsystem_pthread.dylib`_pthread_start + 136
˜˜˜

As until https://github.com/gazebosim/gz-sim/issues/44 is fixed macOS is effectively quite hard to use on macOS, I think it make sense to just skip macOS tests for now, and merge this PR.

xela-95:
> Ok, I fixed the problem by disabling the tests on macOS, we can re-enable them once macOS is actually supported upstream (see [gazebosim/gz-sim#44](https://github.com/gazebosim/gz-sim/issues/44)). At this point, once CI is happy I think we can squash and merge, what do you think?

I agree! Merging!

PR #254, opened 2025-03-18
Do not print error if default refSpeed and refAcceleration are used in gzyarp::ControlBoard plugin

traversaro:
I noticed errors like:

~~~
Error:  refSpeed  not found
Warning:  Parameter refSpeed not found in TRAJECTORY_GENERATION group. Defaults will be applied
Error:  refAcceleration  not found
Warning:  Parameter refAcceleration not found in TRAJECTORY_GENERATION group. Defaults will be applied
~~~

on tests. However, `refSpeed` and `refAcceleration` are not required parameters (at least at the moment), so we should not print any error. as anyhow we already print the warning.

### DIFF
--- a/plugins/controlboard/include/ControlBoard.hh
+++ b/plugins/controlboard/include/ControlBoard.hh
@@ -86,7 +86,8 @@ private:
                      std::vector<double>& out,
                      const std::string& key,
                      const std::string& txt,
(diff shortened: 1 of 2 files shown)


xela-95 (approved):
Thanks! LGTM

PR #258, opened 2025-04-01
Add workaround for compilation with apt dependencies with CMake >= 4.0.0

traversaro:
Without this fix, the apt CI jobs are failing with:

~~~
-- Found Python3: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/jsoncpp/jsoncppConfig.cmake:2 (cmake_policy):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
Call Stack (most recent call first):
  /usr/share/cmake/gz-cmake3/cmake3/FindJSONCPP.cmake:34 (find_package)
  /usr/lib/x86_64-linux-gnu/cmake/gz-fuel_tools9/gz-fuel_tools9-config.cmake:94 (find_package)
  /usr/lib/x86_64-linux-gnu/cmake/gz-sim8/gz-sim8-config.cmake:97 (find_package)
  CMakeLists.txt:17 (find_package)
~~~

This happened as CMake was upgraded (indipendently from apt) to CMake 4.0.0 in the GitHub Actions image, but the `/usr/lib/x86_64-linux-gnu/cmake/jsoncpp/jsoncppConfig.cmake` has a `cmake_minimum_required` call that is not compatible with CMake 4.0.0 . Setting `CMAKE_POLICY_VERSION_MINIMUM` env variable to `3.5` should solve this problem. No documentation is necessary for this, as most apt users will not install a recent CMake, so they are not subject to this problem.

### DIFF
--- a/.github/workflows/apt22.yml
+++ b/.github/workflows/apt22.yml
@@ -11,6 +11,11 @@ on:
   # Execute a "nightly" build at 2 AM UTC
   - cron:  '0 2 * * *'
 
+# Workaround from conflict between GitHub Actions CI upgrading to CMake 4.0.0
+# and jsonCppConfig.cmake file installed by apt that is not compatible with CMake 4.0.0 
+env:
+  CMAKE_POLICY_VERSION_MINIMUM: 3.5
+
 jobs:
   build:
     name: '[${{ matrix.os }}@${{ matrix.build_type }}@yarp:${{ matrix.yarp_tag }}@gazebo:${{ matrix.gazebo_distro }}]'
--- a/.github/workflows/apt22.yml
+++ b/.github/workflows/apt22.yml
@@ -11,11 +11,6 @@ on:
   # Execute a "nightly" build at 2 AM UTC
   - cron:  '0 2 * * *'
 
-# Workaround from conflict between GitHub Actions CI upgrading to CMake 4.0.0
-# and jsonCppConfig.cmake file installed by apt that is not compatible with CMake 4.0.0 
-env:
-  CMAKE_POLICY_VERSION_MINIMUM: 3.5
-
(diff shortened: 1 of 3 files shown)


Nicogene (approved):

traversaro:
Actually I do not think this is a good idea. Even if "normal" users will not use an old apt packages with CMake 4.0.0, anyone building this projects using GitHub Actions will experience this problem, so a solution like https://github.com/pytorch/pytorch/pull/150324 may be a better solution, let me think about this.

traversaro:
> Actually I do not think this is a good idea. Even if "normal" users will not use an old apt packages with CMake 4.0.0, anyone building this projects using GitHub Actions will experience this problem, so a solution like [pytorch/pytorch#150324](https://github.com/pytorch/pytorch/pull/150324) may be a better solution, let me think about this.

More generic fix provided in https://github.com/robotology/gz-sim-yarp-plugins/pull/258/commits/9897f1b770d4b57c992b16e26c14c4fbd89cc2bd .

traversaro:
@xela-95 this is ready for review

xela-95 (approved):
Thanks! 

Just a curiosity: do the placing of the clauses inside the CMakeLists (where setting and unsettling the `CMAKE_POLICY_VERSION_MINIMUM` variable) has a specific rationale?

PR #269, opened 2025-05-30
Export gz-sim-yarp-commons headers

xela-95:
I'm implementing custom gz-sim-yarp-plugins external to this repo, and I noticed we are missing to export the commons library headers. This PR fixes this.


traversaro:
Just a word of warning: I am totally agree on exporting those headers, but please note that now that we export this headers, we have to at least try to respect https://github.com/robotology/gz-sim-yarp-plugins?tab=readme-ov-file#compatibility-policy for those headers, while before this PR we could change everything in those headers without the need to worry if outside code could break. That is perfectly fine, but we should also double check if there is anything we want to keep private.

traversaro:
> It appears to me that https://github.com/robotology/gz-sim-yarp-plugins/blob/main/libraries/common/YarpDevReturnValueCompat.h is used only by the laser plugin and not from common headers. Can I move this header outside the common shared library or can I just continue not exporting it as it's today? or this is a sort of code smell to avoid?

Keeping it in `common` but being actually private is perfectly fine for me.

traversaro (approved):

traversaro:
> > It appears to me that https://github.com/robotology/gz-sim-yarp-plugins/blob/main/libraries/common/YarpDevReturnValueCompat.h is used only by the laser plugin and not from common headers. Can I move this header outside the common shared library or can I just continue not exporting it as it's today? or this is a sort of code smell to avoid?
> 
> Keeping it in `common` but being actually private is perfectly fine for me.

Probably I would just add a comment in the list of the installed headers to explain that `YarpDevReturnValueCompat.h` is intentionally not installed, as it is a private header, and should not be included in any public header.

xela-95:
> Probably I would just add a comment in the list of the installed headers to explain that `YarpDevReturnValueCompat.h` is intentionally not installed, as it is a private header, and should not be included in any public header.

Agree thanks! Done in https://github.com/robotology/gz-sim-yarp-plugins/pull/269/commits/093fa265ad402376315345b6a4ad39e8751ef8ca

xela-95:
Windows action have been cancelled due to #276

xela-95:
Merging!

PR #270, opened 2025-06-11
Fix IMU orientation export

xela-95:
This PR fixes a bug in how the orientation given from a gz-sim IMU sensor is read and exposed via the [IOrientationSensors](https://www.yarp.it/latest/classyarp_1_1dev_1_1IOrientationSensors.html#adb2a40d03c747aa9c8b7fc542f193fd9) interface.

### DIFF
--- a/plugins/imu/Imu.cc
+++ b/plugins/imu/Imu.cc
@@ -1,3 +1,4 @@
+#include <Common.hh>
 #include <ConfigurationHelpers.hh>
 #include <DeviceRegistry.hh>
 #include <ImuDriver.cpp>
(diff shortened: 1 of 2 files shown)


traversaro:
Thanks, do you have any idea on the failures?

xela-95:
> Thanks, do you have any idea on the failures?

Yes it was a unit test testing a wrong assumption on how the orientation should have been:
https://github.com/robotology/gz-sim-yarp-plugins/blob/0875a384fba3d44b377f9c19c90e047f332b9e0c/tests/imu/ImuTest.cc#L117

This was basically masking the fact that the norm of the quaternion was used instead of the yaw angle.

Issue #279, opened 2025-08-29
Update to Yarp4


randaz81:
The branch YARP4 is ready.

traversaro:
> The branch YARP4 is ready.

Thanks @randaz81 ! I adapted it in https://github.com/robotology/gz-sim-yarp-plugins/tree/yarp4support to add support for yarp 4 without breaking backward compatibility with YARP 3.12, so that we can avoid to maintain two different branches.

traversaro:
It is deprecated, but people are still using it, so I think it make sense to just easier to update it, it should be easy to do, I can handle it.

I have just one question, in the YARP 4 changelog (https://github.com/robotology/yarp/blob/8c1b42c7b4fabb777476c95f56968858881c0b61/doc/release/v4_0_0.MD?plain=1#L41) I read "Removed deprecated device `RGBDSensorClient` (replaced by RGBDSensor_nwc_yarp)". I was a bit confused by this working, I guess what happened is that `RGBDSensorClient` was renamed to `RGBDSensor_nwc_yarp`, but the device parameters are the same?

traversaro:
Ack, but given that if I am not wrong the device was never deprecated in the past (i.e., you still need to use `RGBDSensorClient` if you use YARP 3.12), perhaps writing "The `RGBDSensorClient` YARP devices was renamed to `RGBDSensor_nwc_yarp`" may be more clear? At least, it would be for me.

PR #280, opened 2025-08-30
Migrate BaseState plugin to use `MultipleAnalogSensors` interface and update basestate tutorial and unit tests

traversaro:
This change is required as in YARP 4.0.0 the `analogServer` device has been removed. This raises the minimum version required to build `gz-sim-yarp-plugins` to YARP 3.11.0 .

Fix part of https://github.com/robotology/gz-sim-yarp-plugins/issues/279 .


plugins/basestate/BaseStateDriver.cpp
```diff
@@ -114,40 +134,352 @@ class yarp::dev::gzyarp::BaseStateDriver : public yarp::dev::DeviceDriver,
...
 
 private:
-    ::gzyarp::BaseStateData* m_baseLinkData;
+    ::gzyarp::BaseStateData* m_baseLinkData = nullptr;
     std::string m_baseLinkName;
+
+    static inline double rad2deg(double r)
+    {
+        return r * 57.2957795130823208768; // 180 / pi
+    }
```
traversaro, line 484: We should use the existing functions, not add a new one.

xela-95: Done in https://github.com/robotology/gz-sim-yarp-plugins/pull/280/commits/629c9e4a8f4d7796f83985df77646e3eaec67f9b

traversaro:
> Hi @traversaro, thanks for this PR! One quesiton: in the title it mention changes to the `basestate` tutorial but it was actually not updated here. Do you have some unpushed commits? otherwise I can work on it

Indeed yes, that is the case. The fix is trivial, let me just do it locally.

xela-95:
In https://github.com/robotology/gz-sim-yarp-plugins/pull/280/commits/5145ee2e97d6f7d830d3376014d821beaf8e73fe I added some basic unit tests on the basestate plugin. However when I'm running the tutorial (a box falling on the ground under the effect of gravity), at steady state I'm getting an almost zero vertical linear acceleration. should this be instead the proper accelearation that would be measured by an accelerometer instead?

This is one of the measurements I get doing ` yarp read /read ... /basestate/measures:o`:

```
(((0.0 0.0 0.0) 80.8719999999999998863)) (((3.83121999999999986384e-16 2.89850999999999984023e-16 -0.00980318999999999982076) 80.8719999999999998863)) () (((2.19053224234416349613e-16 1.89550481447540020741e-16 -9.28036929507248396005e-20) 80.8719999999999998863)) () () () () () (((1.64933000000000004765e-19 -1.91282999999999988778e-19 0.049903900000000001036) 80.8719999999999998863)) (((-3.31137000000000006466e-19 6.44383000000000032403e-19 0.000970515000000000027831) 80.8719999999999998863)) (((0.0 0.0 0.0) 80.8719999999999998863))
```

xela-95:
Ok now the new tests are passing. It remains only this to be understood: https://github.com/robotology/gz-sim-yarp-plugins/pull/280#issuecomment-3242767083

CC @traversaro

xela-95:
@traversaro I think this is ready to be merged. Let me know if it's ok for you

traversaro:
Thanks, good to go for me!

PR #281, opened 2025-08-30
 Add support for YARP 4.0.0 keeping backward compatibility with YARP 3.12 

### LINKED ISSUE #279: Update to Yarp4

traversaro:
Most of the code was written by GitHub Copilot + GPT-5, with the following prompt:

> Can you fix gz-sim-yarp-plugins to compile with YARP 4.0 ? You can take https://github.com/robotology/gz-sim-yarp-plugins/commit/21577ef0ec6e540ee13e2161202cff35eedd4c63 as an initial point to work, but the problem of that one is that break compatibility with YARP 3.11, while we want to have a single branch that compiles fine with both YARP 3.11, 3.12 and YARP 4 . To do that, you can use the YarpDevReturnValueCompat.h helpers, and take ispiration from how https://github.com/robotology/gz-sim-yarp-plugins/pull/240 added support for YARP 3.12 without breaking the compatibility with YARP 3.11 . Thanks a lot in advance!

The code was then cleaned up in some details manually.

Fix https://github.com/robotology/gz-sim-yarp-plugins/issues/279 .

### DIFF
--- a/libraries/common/gzyarp/YarpDevReturnValueCompat.h
+++ b/libraries/common/gzyarp/YarpDevReturnValueCompat.h
@@ -5,11 +5,17 @@
 #include <yarp/conf/version.h>
 
 // These macros simplify the migration of the devices implementation from YARP 3.11 to YARP 3.12,
+// and from YARP 3.12 to 4.0
 // where the interfaces migrated from using bool as return values to use yarp::dev::ReturnValue,
 // see https://github.com/robotology/yarp/discussions/3168
 //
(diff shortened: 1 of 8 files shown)


xela-95 (approved):
Thanks @traversaro it looks good to me!

xela-95:
I'm looking at #280 right now, if you prefer we can then merge that before and then this one :)

xela-95:
Ok! Merging then!

Issue #267, opened 2025-05-26
Wrong behavior when switching to IDLE control mode and then back to POSITION one


LoreMoretti:
As per this video:

https://github.com/user-attachments/assets/f01f91a4-d969-4282-a0ff-44da44cb1690

It seems that the reference point does not get updated when switching to IDLE control model and then back to the POSITION one. This causes the joint to go back to the previous position, instead of staying at the one reached in IDLE mode.

traversaro:
This is kind of related to https://github.com/robotology/gz-sim-yarp-plugins/issues/97 . Beside the indications there, what we need to do is  to implement the correct setpoint reset logic in https://github.com/robotology/gz-sim-yarp-plugins/blob/main/plugins/controlboard/src/ControlBoardData.cpp#L115 .

Issue #97, opened 2024-02-23
Manage change of control mode in control board


xela-95:
It is of utmost importance to manage correctly the switch from one control mode to another, to guarantee the smoothest transition, for example by properly setting new correct references to track.

xela-95:
@traversaro if there are specifications (written somewhere or from experience on the robot) let me know, so that we can gather them here.

traversaro:
> @traversaro if there are specifications (written somewhere or from experience on the robot) let me know, so that we can gather them here.

I am not aware if there are any written specification, but what happens in most implementation is that the measured quantity is used to initialize the setpoint, see for example:
* https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriverControlMode.cpp#L141
* https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriverControlMode.cpp#L174

Interestingly, this logic seems not be used in velocity, but that is probably a bug:
* https://github.com/robotology/gazebo-yarp-plugins/blob/cb3521f7543a5a07e2a65a82dedd2eb2ec2cea12/plugins/controlboard/src/ControlBoardDriverControlMode.cpp#L157

traversaro:
> Interestingly, this logic seems not be used in velocity, but that is probably a bug:

Thinking a bit more about this, perhaps the logic is to do the safest possible thing in the change of control mode, i.e. ensure that the desired velocity is 0, so this could have sense.

PR #286, opened 2025-09-13
Reset reference values on control mode change

PeterBowman:
This patch aims to prevent residual motion to happen when control mode is changed back and forth, e.g. a position command is executed first, then the joint has been moved using a different control mode, then the initial mode is reinstated (here, the initial reference is applied, causing undesired motion). See linked issues below for context.

Besides the aforementioned fix, a few more changes are included:

- Minor code style improvements and fixed typos. A quite lengthy `if` clause has been turned into a switch.
- `setControlMode` is now a no-op if the requested mode matches the current active mode.

Note: an inline comment states that "The only not supported control mode is (for now) VOCAB_CM_MIXED", but it is allowed to switch into this mode regardless. I was unable to track the original code beyond https://github.com/robotology/gz-sim-yarp-plugins/commit/b9c56cbe8c4a905997a5f8192f6423a58f965f5e#diff-7b78d4fa28c11f70eb0d5d0c6978eb8ec0d9ddeae84d77e65d5569f763ba21b7R244-R255.

Resolves https://github.com/robotology/gz-sim-yarp-plugins/issues/97.
Resolves https://github.com/robotology/gz-sim-yarp-plugins/issues/267.

### DIFF
--- a/plugins/controlboard/src/ControlBoardData.cpp
+++ b/plugins/controlboard/src/ControlBoardData.cpp
@@ -52,7 +52,8 @@ bool ControlBoardData::setInteractionMode(int axis, yarp::dev::InteractionModeEn
     return true;
 }
 
-bool ControlBoardData::setControlMode(int j, int mode) {
+bool ControlBoardData::setControlMode(int j, int mode)
+{
     int desired_mode = mode;
 
     if (j < 0 || j >= this->actuatedAxes.size())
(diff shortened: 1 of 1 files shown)


xela-95:
Thanks for your PR @PeterBowman !! We'll review it as soon as possible! Probably, we also need to add a unit test to fix the desired behavior.

PeterBowman:
Hello, @xela-95. I have added a new test case to ControlBoardPositionControlTest. Logic:

1. Move to a pose A in POSITION mode.
2. Switch to POSITION_DIRECT and go to pose B.
3. Switch again to POSITION mode.
4. Verify that the robot stays at pose B.

Before this patch, this test fails on the last step (robot at pose A).

xela-95 (approved):

PR #292, opened 2025-10-06
Add find_dependency(YARP COMPONENTS robotinterface os) in gz-sim-yarp-plugins CMake config

traversaro:
Without thix, `find_package(gz-sim-yarp-plugins)` and then using the `gz-sim-yarp-plugins::gz-sim-yarp-commons` target fails with:

~~~
 │ -- Configuring done (7.5s)
 │ CMake Error at $SRC_DIR_run_env/lib/cmake/gz-sim-yarp-plugins/gz-sim-yarp-pluginsTargets.cmake:61 (set_target_properties):
 │   The link interface of target "gz-sim-yarp-plugins::gz-sim-yarp-commons"
 │   contains:
 │     YARP::YARP_robotinterface
 │   but the target was not found.  Possible reasons include:
 │     * There is a typo in the target name.
 │     * A find_package call is missing for an IMPORTED target.
 │     * An ALIAS target is missing.
 │ Call Stack (most recent call first):
 │   $SRC_DIR_run_env/lib/cmake/gz-sim-yarp-plugins/gz-sim-yarp-pluginsConfig.cmake:28 (include)
 │   CMakeLists.txt:5 (find_package)
~~~

### DIFF
--- a/.github/workflows/conda-forge.yml
+++ b/.github/workflows/conda-forge.yml
@@ -102,4 +102,4 @@ jobs:
       run: |
         cd build
         export CMAKE_PREFIX_PATH=${GITHUB_WORKSPACE}/install
-        cmake-package-check gz-sim-yarp-plugins --targets gz-sim-yarp-plugins::gz-sim-yarp-device-registry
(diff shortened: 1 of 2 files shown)


xela-95 (approved):
Thanks a lot @traversaro !!

PR #304, opened 2026-01-21
worldinterface plugin [MVP]

randaz81:
added new plugin `worldinterface` to add/move/delete models in gazebo.

This PR is a MVP, for preliminary testing only.
Some functionalities still need to be implemented in upcoming PRs.


Nicogene (changes requested):
Thank you @randaz81 !

PR #307, opened 2026-01-28
Add YARP 4.0.0 support for controlboard plugin

martinaxgloria:
Starting from:

- https://github.com/robotology/gz-sim-yarp-plugins/pull/281

The goal is to have complete support for yarp4 but keeping backward compatibility with previous yarp versions.

cc @valegagge

### DIFF
--- a/plugins/controlboard/include/ControlBoardDriver.hh
+++ b/plugins/controlboard/include/ControlBoardDriver.hh
@@ -3,6 +3,7 @@
 #include <ControlBoardData.hh>
 
 #include <string>
+#include <vector>
 
 #include <yarp/dev/DeviceDriver.h>
 #include <yarp/dev/IAxisInfo.h>
@@ -19,6 +20,7 @@
 #include <yarp/dev/IVelocityControl.h>
 #include <yarp/dev/PidEnums.h>
 #include <yarp/os/Searchable.h>
+#include <gzyarp/YarpDevReturnValueCompat.h>
 
 namespace yarp
 {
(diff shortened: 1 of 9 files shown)


martinaxgloria:
To test the changes made with this PR, I think that it would be nice to run the tests included in the CI locally with `yarp4` installed on my computer, except for the one that requires icub-main (I set `GZ_SIM_YARP_PLUGINS_ENABLE_TESTS_WITH_ICUB_MAIN` to OFF). For the backward compatibility, instead, I think we can rely on the CI. 

I built the tests and I ran them following the instructions given by the CI. They all passed except for the `DepthCameraTest`: this is strange since I didn't touch the DepthCamera plugin, nor the test, and within the CI the test passes.

Moreover, I checked an iCub model on gz-sim to see if the controlboard plugins start and close without any errors.

martinaxgloria:
> The CI is happy, do you mean locally?

Yes, I ran all the tests locally to check if they pass also with yarp4 (the CI relies on yarp-3.12). To narrow down the problem, I installed yarp-3.12, I re-ran the DepthCamera test but I failed also in this case. Probably, there are some issues in my local configuration that do not exactly match the one in which the CI runs

martinaxgloria:
Looking more in depth at the output of the test, I saw that when the device [rgbdSensor_nws_yarp](https://github.com/robotology/gz-sim-yarp-plugins/blob/b68a4d69f7f88149bd9a192436f5e24d9850e1a4/tests/depthcamera/depthcamera_nws.xml#L6) was opened, it was unable to read the proper value of the parameter `name`:

```
[INFO] |yarp.device.RgbdSensor_nws_yarp| Parameter 'period' using value: 0.033333
[INFO] |yarp.device.RgbdSensor_nws_yarp| Parameter 'name' using value: name
[ERROR] |yarp.os.Port| Port name 'name/rpc:i' needs to start with a '/' character
[ERROR] |yarp.devices.RgbdSensor_nws_yarp| Unable to open rpc Port name/rpc:i
[ERROR] |yarp.os.Port| Port name 'name/rgbImage:o' needs to start with a '/' character
[ERROR] |yarp.devices.RgbdSensor_nws_yarp| Unable to open color streaming Port name/rgbImage:o
[ERROR] |yarp.os.Port| Port name 'name/depthImage:o' needs to start with a '/' character
[ERROR] |yarp.devices.RgbdSensor_nws_yarp| Unable to open depth streaming Port name/depthImage:o
[INFO] |yarp.dev.PolyDriver|depthcamera_nws_yarp| Created wrapper <rgbdSensor_nws_yarp>. See C++ class RgbdSensor_nws_yarp for documentation.
```

I don't know if it is something related to yarp4, but the correct device name is `RGBDSensor_nws_yarp`. In fact, changing the device name, at least I'm able to open the device.

martinaxgloria:
Another proposal could be replace `depthcamera_nws.xml` with `depthcamera_nws.xml.in` with `@DEPTH_CAMERA_DEVICE_NAME@` in place of the device type. 

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE robot PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">

<robot name="camera" portprefix="camera" build="0"  xmlns:xi="http://www.w3.org/2001/XInclude">
    <devices>
        <device name="depthcamera_nws_yarp" type="@DEPTH_CAMERA_DEVICE_NAME@">
        <param name="name"> /depthcamera </param>
            <param name="period"> 0.033333 </param>
            <action phase="startup" level="5" type="attach">
                <!-- This is the same name that we passed with the yarpDeviceName to the camera plugin -->
                <param name="device"> depthcamera_plugin_device </param>
            </action>
            <action phase="shutdown" level="5" type="detach" />
        </device>
    </devices>
</robot>
```

Then, in the CMakeLists of the test, add:

```cmake
if(YARP_VERSION VERSION_GREATER_EQUAL "4.0.0")
  set(DEPTH_CAMERA_DEVICE_NAME "RGBDSensor_nws_yarp")
else()
  set(DEPTH_CAMERA_DEVICE_NAME "rgbdSensor_nws_yarp")
endif()

configure_file(depthcamera_nws.xml.in depthcamera_nws.xml @ONLY)
```

In this way, `depthcamera_nws.xml` is generated, replacing the placeholder with the correct device type

Nicogene (approved):

randaz81 (changes requested):
ControlBoardDriver.cpp implementation to be fixed for new controlmodes and pidtypes

valegagge:
> Even if the tests are ok, there are large sections of the controlBoard plugin that need to be rewritten (the CI is ok only because the related tests are missing). The errors will appear, however, if the user tries to access the robot interfaces, for example, with yarpmotorgui (which tries to get all the capabilities of the simulated board). The current implementation misses the new types of PIDs introduced in Y4, their parsers, their handles (set/getPid) the new control modes (set/getControlModes) etc. (see fakeMotionControl and embObjMotion). I can take care of this.

Hi @randaz81 , 
how is this activity going? 😸

randaz81:
I now implemeted all the required changes. I havent retested the gazebo plugins, though.

Nicogene:
This branch needs to be rebased on latest `main`

martinaxgloria:
Hi @Nicogene, I rebased my branch on the last main

cc @randaz81

Nicogene (approved):

PR #316, opened 2026-05-23
Cache gz::sim::Joint to improve performance in ControlBoard plugin

PeterBowman:
While testing a 70-joint robot (UC3M RoboticsLab's TEO humanoid, including fixed joints), I found there is room for improvement by avoiding to query and build `gz::sim::Joint` instances from joint names on every simulation update cycle. This patch caches said instances, and incidentally also avoids calling `std::vector::at()` in favor of the bracket notation due to performance concerns as well ([SO](https://stackoverflow.com/a/6911339)), and since there are no surrounding `try-catch` blocks anyway (except for one occurrence I decided to leave as-is).

On my setup (Intel Core i7-10700F), performance was improved by a 20% (the real-time factor displayed on Gazebo Sim's lower right corner raised from 50% to 60%).

### DIFF
--- a/plugins/controlboard/src/ControlBoard.cpp
+++ b/plugins/controlboard/src/ControlBoard.cpp
@@ -107,7 +107,7 @@ void ControlBoard::Configure(const Entity& _entity,
     std::string yarpDeviceName = m_pluginParameters.find("yarpDeviceName").asString();
 
     m_robotScopedName = gz::sim::scopedName(_entity, _ecm, "/");
-    m_modelEntity = _entity;
+    m_model = Model(_entity);
 
     m_pluginParameters.put(yarp::dev::gzyarp::YarpControlBoardScopedName.c_str(),
                            m_robotScopedName.c_str());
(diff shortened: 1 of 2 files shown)


plugins/controlboard/src/ControlBoard.cpp
```diff
@@ -524,8 +513,10 @@ bool ControlBoard::updateReferences(const UpdateInfo& _info, EntityComponentMana
         }
     }
 
-    for (auto& joint : m_controlBoardData.physicalJoints)
+    for (int i = 0; i < m_controlBoardData.physicalJoints.size(); i++)
```
Nicogene, line 516: Thanks @PeterBowman for this improvement!

Why did you need to make this change?

Nicogene (approved):
Super @PeterBowman!

Just one small doubt

plugins/controlboard/src/ControlBoard.cpp
PeterBowman: Hi, @Nicogene! Further down I needed to access `m_gzJoints[i]`, therefore the for-each loop has been turned into a for-i.

PR #226, opened 2024-12-20
Add `DepthCamera` gz-sim plugin

Nicogene:
It implements the `gzyarp::DepthCamera` device, that exposes both RGB and depth images via the [`yarp::dev::IRGBDSensor` ](https://www.yarp.it/v3.4/classyarp_1_1dev_1_1IRGBDSensor.html) interface.

It fixes #106

PR #317, opened 2026-05-25
Fmarta/multicameraplugin

fincatomarta:
Added a new plugin called **stereo camera** which launches two cameras at once, one simulating the right eye one simulated the left eye.


Nicogene (changes requested):
Hello @fincatomarta, thanks for the nice addition!

Could you please add a unit test for testing this new plugin?
It would be nice to have also an example of usage under the `tutorial` folder

You can use this PR as reference:
- https://github.com/robotology/gz-sim-yarp-plugins/pull/226

Nicogene (approved):
Thank you @fincatomarta, LGTM!

PR #322, opened 2026-07-01
Add joint type handling for prismatic and revolute joints

kaidegast:
## Summary
This PR adds joint-type-aware behavior to the ControlBoard plugin so prismatic joints are handled correctly (without angular unit conversions), while revolute joints keep the existing degree/radian conversion flow.

link to the comment detailing the issue encountered: https://github.com/mesh-iit/mobile-base/issues/197#issuecomment-4752829564

## Changes
1. Added joint type storage to per-joint physical properties.
2. Detected joint type from Gazebo joint components during plugin setup.
3. Updated conversion helpers to branch by joint type:
   - Prismatic joints: identity conversion (no degree/radian transforms).
   - Revolute joints: existing degree/radian and gain conversions.
4. Updated `getJointType(...)` to:
   - Return the stored joint type instead of always returning revolute.
   - Guard access with a mutex.
   - Validate axis bounds and report an error on invalid indices.
5. Added required YARP axis/joint type include for joint-type API support.

## Behavior Impact
- Fixes unit-conversion behavior for prismatic joints.
- Improves correctness of joint type reporting through the controlboard driver.
- Keeps revolute-joint behavior unchanged.

## Compatibility
- Backward compatible for existing revolute-only setups.
- No interface-breaking changes expected.

### DIFF
--- a/plugins/controlboard/src/ControlBoard.cpp
+++ b/plugins/controlboard/src/ControlBoard.cpp
@@ -272,10 +272,16 @@ bool ControlBoard::setJointProperties(EntityComponentManager& _ecm)
                 m_controlBoardData.physicalJoints[i].jointType =
                     yarp::dev::JointTypeEnum::VOCAB_JOINTTYPE_PRISMATIC;
             }
-            else
+            else if (jointTypeComponent != nullptr && jointTypeComponent->Data() == sdf::JointType::REVOLUTE)
             {
                 m_controlBoardData.physicalJoints[i].jointType =
                     yarp::dev::JointTypeEnum::VOCAB_JOINTTYPE_REVOLUTE;
+            } 
+            else 
+            {
+                yError() << "Joint " << jointFromConfigName
+                         << " has unsupported joint type. Only PRISMATIC and REVOLUTE are supported.";
+                return false;
             }
 
             // Initialize JointProperties object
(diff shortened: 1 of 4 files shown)


plugins/controlboard/src/ControlBoard.cpp
```diff
@@ -263,6 +265,19 @@ bool ControlBoard::setJointProperties(EntityComponentManager& _ecm)
...
             gzJoint.EnableTransmittedWrenchCheck(_ecm, true);
 
+            auto jointTypeComponent = _ecm.Component<gz::sim::components::JointType>(jointEntity);
+
+            if (jointTypeComponent != nullptr && jointTypeComponent->Data() == sdf::JointType::PRISMATIC)
+            {
+                m_controlBoardData.physicalJoints[i].jointType =
+                    yarp::dev::JointTypeEnum::VOCAB_JOINTTYPE_PRISMATIC;
+            }
+            else
```
Nicogene: I'd prefere to make the if explicit for the revolute joint, otherwise if the fallback is revolute we risk that some joints (e.g. fixed) are treated as revolute

Nicogene (changes requested):
One small request

plugins/controlboard/src/ControlBoard.cpp
kaidegast: Added handling for unsupported joint types. This way only revolute and prismatic are allowed. Let me know if this changes are satisfactory or if the unspecified joints should be handled differently.

Nicogene (approved):
I tested ergoCub on my setup and it works flawlessly, thank you @kaidegast !

